PR #36, opened 2014-03-26
Update 07_resource_types_and_traits.md

mmontag:
"schemas property must be an array."


sichvoge:

PR #41, opened 2014-06-09
Remove redundant `the`

Philzen:
Also i have a question about the same chapter where it says in the first section:

> A trait is a partial method definition that, like a method, can provide method-level properties such as description, headers, query string parameters, and responses. _Methods that use one or more traits inherit those traits' properties._

and -  hardly two paragraphs further, in the middle of resource type explanation, it merely repeats:

> _Similarly, methods may specify one or more traits from which they inherit using the is property._

Just an observation, believe this could be streamlined somehow.

### DIFF
--- a/07_resource_types_and_traits.md
+++ b/07_resource_types_and_traits.md
@@ -42,7 +42,7 @@ traits:
           required: true
 ```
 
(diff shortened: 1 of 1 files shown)


sichvoge:

PR #89, opened 2014-09-09
Update raml-0.8.md

vange:
/user instead of /users typo


sichvoge:
This has been fixed for raml-0.8.md, at least in branch `raml-10-rc2`.

PR #90, opened 2014-09-12
The spec says schemas should be an array of maps

blakeembrey:


sichvoge:

PR #144, opened 2015-03-16
Fixed value of required field

ddossot:
Used true instead of yes, per: https://github.com/raml-org/raml-spec/blob/master/raml-0.8.md#required


sichvoge:

PR #155, opened 2015-07-08
Gh pages change

jbnicolai:
To me 'with a `!includes` sounds more natural and 'an', but I chose to drop the a/an altogether as that seems to be done more frequently, e.g. lines 23 and 24.

The pull-request is from, and targets, gh-pages-staging as I could find no documentation on where to introduce changes regarding the website.


jbnicolai:
Apologies, forgot to switch to `gh-pages-staging`...

PR #158, opened 2015-07-09
Fixing typos

cerivera:
Fixing a simple typo and also fixing invalid paths according to the example RAML spec that they're referencing.


sichvoge:
Will be fixed with RC2.

PR #210, opened 2015-12-11
Proposal for specification of Cookies. Inspired by `headers`.

vanDonselaar:


sichvoge:
Hi @vanDonselaar, I'd like to rather have an issue about introducing something new rather than putting it directly into the spec yet. That we can make sure to discuss and agree. Please see the https://github.com/raml-org/raml-spec/tree/raml-10#contributing section for more information.

vanDonselaar:
No problem: https://github.com/raml-org/raml-spec/issues/484

PR #211, opened 2015-12-11
Add UUID type.

vanDonselaar:


raml-0.8.md
```diff
@@ -156,6 +156,7 @@ The *type* attribute specifies the primitive type of the parameter's resolved va
 | integer | Value MUST be an integer. Floating point numbers are not allowed. The integer type is a subset of the number type.
 | date    | Value MUST be a string representation of a date as defined in RFC2616 Section 3.3 [RFC2616]. See [Date Representations](#date-representations).
 | boolean | Value MUST be either the string "true" or "false" (without the quotes).
+| uuid    | Value MUST be a Universally unique identifier (UUID). E.g. "f47ac10b-58cc-4372-a567-0e02b2c3d479".
```
dmartinezg: Is it worth it to point to the RFC? (https://tools.ietf.org/html/rfc4122 SeEction 3, provides a strict grammar, which can be used by parsers to validate)

vanDonselaar: Good idea. Thanks. PR updated.

sichvoge:
Hi @vanDonselaar, I'd like to rather have an issue about introducing a new type than putting it directly into the spec yet. That we can make sure to discuss and agree. Please see the https://github.com/raml-org/raml-spec/tree/raml-10#contributing section for more information.

vanDonselaar:
Sure. See: https://github.com/raml-org/raml-spec/issues/483

PR #263, opened 2016-02-29
this PR closes issue 251 - clarification about traits, resource types, and security schemes

### LINKED ISSUE #251: Spec is not really clear about behaviour of securedBy on resources
Are security schemes applied to a resource also automatically applied to all nested resources?

sichvoge:
clarify how traits, resource types, and security schemes behave on nested resources. closes #251


usarid:
LGTM

PR #265, opened 2016-03-01
modified template uris from level 1 to also support level 2

### LINKED ISSUE #73: Allow '/' in URI parameters - support level 2 URI templates
The spec says:

>  The baseUri property's value MUST conform to the URI specification [RFC2396] or a Level 1 Template URI as defined in RFC 6570 [RFC6570].

In order to be able to document more versatile APIs, _Level 2_ templates should also be allowed, but only if the macro is the last piece in the URL

Example:

```yaml
#%RAML 0.8
title: My Sample API
/files{+path}:
  uriParameters:
    path:
      description: The full path to the desired file
      type: string
      example: /my/file
```

sichvoge:
The main effort is to close #73, but I also introduced a global definition for "Template URIs" inside Definitions and Terminology chapter to avoid any redundancy. I also added a few more that can be linked throughout the document later.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -19,15 +19,29 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
 
 ## Definitions and Terminology
 
-For this specification, **API definition** will be used to denote the description of an API using this specification and **RAML Specification** refers to the current document.
+### General
 
-**REST** is used in the context of an API implemented using some or all of the principles of REST. The REST acronym stands for Representational State Transfer and was first introduced and defined in 2000 by Roy Fielding in his doctoral dissertation [REST].
+For this specification, **API definition** is used to denote the description of an API using this specification and **RAML Specification** refers to the current document.
+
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -19,16 +19,30 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
...
 
+### Markdown
+
+Throughout this specification, **Markdown** means [GitHub-Flavored Markdown](https://help.github.com/articles/github-flavored-markdown/).
+
+### Media Types
+
+The value of a specific media type property MUST conform to the media type specification in [RFC6838](https://tools.ietf.org/html/rfc6838).
+
+### Template URI
```
sichvoge: @usarid thats the main section for the change; the beauty now is that we can change the text and don't have to look inside the whole document anymore for what we need to change as well. That is the same for markdown and media types therefore added them as well.

usarid:
Let's please separate the PR for the level 1 --> 2 change from the PR for other stuff.

sichvoge:
Shall I still use a global chapter to define what a URI template looks like? Just to make sure, if we don't then the info will be in multiple places across the spec. It's easier to have one place for future updates and linking everything to it.

PR #274, opened 2016-03-07
Clarification about resource types and optional properties

### LINKED ISSUE #143: Parameters for optional Resource Type properties should also be optional
The spec currently states:

_The definitions of resource types and traits MAY contain parameters, whose values MUST be specified when applying the resource type or trait ..._

This means that Resource Type parameters are currently mandatory, even if they're not being used.  Given a snippet like this:

```yaml
# ...
resourceTypes:
  - Resource:
      get:
        description: Get one <<ResourceName>> record.
      patch?:
        description: Some info about <<TextAboutPatch>>.
# ...
```

In the case of a resource with no `patch` support, I still have to provide a blank `TextAboutPatch` property:

```yaml
/MyThing
  type:
    Resource:
      ResourceName: MyThing
      TextAboutPatch:
  get:
    description: Override the default Resource description here.
```

It would be nice if the spec would allow for parameters to be optional if their corresponding resource type properties are unused for the given resource.

sichvoge:
If someone does not explicitly define the optional property in a resource or method, a processor MUST not force one to define the corresponding custom parameter. Closes #143

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -2020,7 +2020,7 @@ When defining resource types and traits, it can be useful to capture patterns th
 
 To accommodate this need, a resource type or trait definition MAY append a question mark ("?") suffix to the name of any object-valued property that should not be applied if it doesn't already exist in the resource or method at the corresponding level. A property is considered object-valued if its value is of type object or of a type that has object at the base of its inheritance hierarchy. The question mark suffix indicates that the value of the property in the resource type or trait should be applied if the property name itself (without the question mark) is already defined (whether explicitly or implicitly) at the corresponding level in that resource or method.
 
-The following example shows a resource type called corpResource with an optional post? property that defines a required header called X-Chargeback. If the inheriting resource defines a post method, it will include the X-Chargeback header requirement. If the inheriting resource does not define a post method, one will not be defined for it by dint of inheriting from the corpResource resource type.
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -2020,7 +2020,7 @@ When defining resource types and traits, it can be useful to capture patterns th
...
 
-The following example shows a resource type called corpResource with an optional post? property that defines a required header called X-Chargeback. If the inheriting resource defines a post method, it will include the X-Chargeback header requirement. If the inheriting resource does not define a post method, one will not be defined for it by dint of inheriting from the corpResource resource type.
+The following example shows a resource type called corpResource with an optional post? property that defines a required header called X-Chargeback and also a custom parameter called TextAboutPost. If the inheriting resource defines a post method, it will include the X-Chargeback header requirement and TextAboutPost MUST be defined as well. If the inheriting resource does not define a post method, one will not have to define the X-Chargeback header or forced to define the TextAboutPost parameter by dint of inheriting from the corpResource resource type.
```
usarid: "...or forced..." --> "...or be forced..."

versions/raml-10/raml-10.md
```diff
@@ -2028,17 +2028,21 @@ title: Example of Optional Properties
...
+  type:
+    corpResource:
+      TextAboutPost: post method # post defined which will force to define the TextAboutPost parameter
   get:
   post: # will require the X-Chargeback header
 /queues:
   type: corpResource
   get:
-  # will not have a post method defined
+  # will not have a post method defined which means one MUST not have to define
```
usarid: Not sure about capitalizing the MUST. You can just say the TextAboutPost parameter is not required because it is in fact not used. I would even ask: can you even submit a TextAboutPost parameter value at all, since it's not used?

sichvoge: Agree - this is the intent right; post has been not used so there is no need to define TextAboutPost

sichvoge: I'd say that we basically state "it's not required" only and be open about how tooling will process it when it is defined. Either it's ignoring it or add a warning. If we need to be make that more clear in the future, we can still change it.

PR #275, opened 2016-03-07
clarify collapsing of consecutive slashes due to a trailing slash in …

### LINKED ISSUE #256: Make explicit the collapsing of consecutive slashes due to a trailing slash in the base URI
When the base URI ends in one or more slashes (`/`), those trailing slashes are to be omitted in constructing the absolute paths for the resources using that base URI. 

For example, in
```raml
baseUri: http://api.test.com/common/
/users:
  /groups:
```
the absolute paths for the two resources are `http://api.test.com/common/users` and `http://api.test.com/common/users/groups`.

In a more complicated examples with consecutive slashes in multiple places, note that only trailing slashes in the base URI are collapsed:

```raml
baseUri: //api.test.com//common//
/:
  /users/:
    /groups//:
```
leads to the following 3 resource absolute paths: `//api.test.com//common/`, `//api.test.com//common//users/` and `//api.test.com//common//users//groups//`.

Various clients and servers may choose to handle these in different ways, but from an HTTP and RAML perspective, these are the resource paths.

sichvoge:
…the base URI. closes #256.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -214,6 +214,28 @@ baseUriParameters:
     description: The name of the bucket
 ```
 
+When the base URI ends in one or more slashes (`/`), those trailing slashes are to be omitted in constructing the absolute paths for the resources using that base URI.
+
+For example, in
+
+```yaml
+baseUri: http://api.test.com/common/
+/users:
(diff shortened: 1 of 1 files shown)


usarid:
LGTM

Issue #202, opened 2015-11-25
date format:  RFC 2616 vs. RFC 3339


gardea:
Is there a particular reason why RFC 2616 was recommended over RFC 3339? I understand the differences between the two; however, there's a strong community believing that ISO 8601 derived formats (RFC 3339 being one) should be the only format used. It would also promote wire level interoperability with other stacks, which at this time seems to predominantly favour 3339...

sichvoge:
There is no particular reason. We have seen some buzz around that ISO format and we can still easily change it to RFC 3339. Would that be better?

https://github.com/raml-org/raml-spec/issues/15

gardea:
I personally consider it would be much better. I also have to disclose that I am heavily influenced by current (large scale) deployments of RAML that we are involved in, the tooling we've built, and our interoperability requirements (XSD to RAML/JSON Schema - XSD is based on ISO8601; RAML to and from Swagger 2.0).

sichvoge:
Would it be better to change it to ISO8601 or RFC3339? (sorry if that is the same, but I understand it is not. RFC3339 is just a kind of ISO8601, right?)

gardea:
You are correct: RFC 3339 is a "simplification" of ISO 8601. My vote goes to RFC 3339.

robertlevy:
+1

sichvoge:
ok cool - accepted. we will change it to RFC 3339

mj1856:
RFC3339 is great, for *timestamps*.  However, it does not cover date-only, time-only, or datetime-without-timezone-offset values.  Those are all covered by ISO 8601.  RAML should specify the ISO8601 *extended format* in general, and can further clarify conformance to RFC3339 for timestamps.

This is relevant, considering #105.

mj1856:
And RFC 822 / 2816 / 2822, et. al. is definitely the wrong choice.  All modern APIs are gravitating to ISO8601 / RFC3339, for many good reasons.

usarid:
We cannot just change to RFC3339 without giving some thought to current users of RAML 0.8 and moreover to users of HTTP :-)

For example, a user of RAML 0.8 that had described a header field as being of type `date` upgrades to 1.0, and now what? How do they migrate? Their spec is now valid 1.0 but doesn't describe their API, as the header values are now invalid. What can they do?

[...]

PR #277, opened 2016-03-07
changed date type representation according to issue 202

sichvoge:
Made modifications to the Date chapter according to our discussion in #202

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -710,12 +710,36 @@ types:
 
 ##### Date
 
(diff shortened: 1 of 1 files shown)


usarid:
LGTM. I didn't check whether there are other places in the spec that need to change. All these types are scalar types, and they have no common base type any more than `number` and `string` have a common base type.

sichvoge:
I will do another check before merging just to be sure.

sichvoge:
Checked - there is no other `date` mentioned in any examples or text.

PR #282, opened 2016-03-08
add support for multiple default media types

### LINKED ISSUE #234: Allow defining multiple default media types
Defining a default media type has been described in RAML 0.8 as a mechanism to define the default media type returned by API responses, and expected from API requests that accept a body. 

The proposal is to extend that to be able to define a list of default media types with the premise that all of them are described by exactly the same types and /or examples.

```yaml
#%RAML 1.0
title: New API
mediaType: [ application/json, application/xml]
types:
  Person:
/list:
  get:
    responses:
      200:
        body: Person[]
/send:
  post:
    body: Person
```

The equivalent RAML to the example above is:

```yaml
#%RAML 1.0
title: New API
types:
  Person:
/list:
  get:
    responses:
      200:
        body:
          application/json: Person[]
          application/xml: Person[]
/send:
  post:
    body:
      application/json: Person
      application/xml: Person
```

If you have cases where you need to override the default behaviour, you MAY define media types explicitly when describing your API response or request. The following example illustrates that:

```yaml
#%RAML 1.0
title: New API
mediaType: [ application/json, application/xml]
types:
  Person:
  Another:
/list:
  get:
    responses:
      200:
        body: Person[]
/send:
  post:
    body:
      application/json:
        type: Another
```

sichvoge:
This PR closes #234.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -260,19 +260,36 @@ baseUri: https://na1.salesforce.com/services/data/{version}/chatter
 
 The media types expected from API requests that contain a body, and returned by API responses, can be defaulted by specifying the OPTIONAL **mediaType** property, so they do not need to be specified within every body definition.
 
-The value of the mediaType property MUST be a media type string conforming to the media type specification in [RFC6838](https://tools.ietf.org/html/rfc6838).
+The value of the mediaType property MUST be a sequence of media type strings conforming to the media type specification in [RFC6838](https://tools.ietf.org/html/rfc6838); with the premise that all of them are described by exactly the same types and/or examples.
 
-This example shows a RAML snippet for an API that accepts and returns JSON-formatted bodies. If the remainder of this API's specification doesn't explicitly specify another media type, this API only accepts and returns JSON-formatted bodies.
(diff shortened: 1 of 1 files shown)


usarid:
Hmm, now that I see it in writing: I'm wary of losing the simplicity you get in the majority of APIs that have only one media type, where we had the simple

```raml
mediaType: application/json
```

and we're now throwing that away and replacing it with

```raml
mediaTypes: [ application/json ]
```

just to accommodate the relatively rare cases of multiple media types. Perhaps we should keep `mediaType` as it was before, and just add `mediaTypes` as a (mutually-exclusive) option that takes an array? Most of the time you won't need it, occasionally you will and it's there for you.

sichvoge:
I don't think we should add another node that ultimately means the same compared to a single item array. We should really restrict the cases introducing mutual-exclusive where ever possible. In this case it would not really add too many advantages to say - let's go for it. I don't see any problems with `mediaTypes: [ application/json ]` and so for all others that basically have a sequence as value. The increase of "simplicity" is not too high here.

usarid:
Recognizing that we had our discussion (and it was a good one with valid points on both sides), let's go with keeping `mediaType` as it was before, and allowing it to take an array and not just a single value -- so that's the only change vs RAML 0.8.

PR #284, opened 2016-03-08
added new chapter that illustrates how one annotates scalar-value nod…

### LINKED ISSUE #225: Expand annotations to allow annotating scalar-valued nodes
It is often useful to annotate scalar-valued nodes, e.g. `baseUri`. Since annotations are applied as extra key-value pairs to nodes that accept key-value pairs already (i.e. map-valued nodes), they cannot currently be applied to scalar-valued nodes. 

For the scope of this proposal, scalar-valued nodes are ones that cannot already be expressed as maps, e.g. the `type` node can take a scalar value or a map value, so it is not covered in this proposal.

The proposal is to allow any scalar-valued node, which is normally expressed as
```yaml
baseUri: http://www.example.com/api
``` 
to alternatively be expressed as a map, with the single allowed key `value`:
```yaml
baseUri:
  value: http://www.example.com/api
```
and then annotations may be applied as usual, e.g.
```yaml
baseUri:
  value: http://www.example.com/api
  (redirectable): true
```

As part of this proposal, the list of `allowedTargets` for annotation type declarations is expanded to include all scalar-valued nodes, such as `baseUri`, `description`, `type`, etc. The spec need not list them explicitly in the `allowedTargets` table, but may refer to a detailed table of all scalar-valued nodes.

sichvoge:
…es and the available scalar-value nodes, and also added Scalar to the annotation target table. This PR will close #225.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -2682,6 +2682,72 @@ annotationTypes:
   (testHarness): usersTest
 ```
 
+#### Annotating Scalar-valued Nodes
+
+It is often useful to annotate scalar-valued nodes, e.g. `baseUri`. Since annotations are applied as extra key-value pairs to nodes that accept key-value pairs already (i.e. map-valued nodes), they cannot be easily applied to scalar-valued nodes. To apply annotations to any scalar-valued node, a RAML processor MUST also support scalar-valued nodes to be expressed as a map, with the single allowed key `value`, as an alternative to the normal syntax.
+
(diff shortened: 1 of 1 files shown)


usarid:
LGTM (except of course that I have not verified that the list of scalar nodes is canonical and that they all make sense)

sichvoge:
I went through couple of iteration with @petrochenko-pavel-a to make sure that the list is complete and the scalars make sense.

PR #289, opened 2016-03-10
removed all occurrences for 'allowMultiple' from the specification

### LINKED ISSUE #288: Remove `allowMultiple` from annotation types
According to the YAML spec, multiple keys with the same name are not allowed. RAML should also comply to that and therefore the spec should not allow to define annotations multiple times to the same node. Hence remove the property `allowMultiple` from the annotation type node.

sichvoge:
closes #288


usarid:
LGTM

Issue #239, opened 2016-02-05
Suggestion to change how you define multiple examples


sichvoge:
In the current RAML RC1 you can define multiple examples either using an array of `-content` keys or unique names followed by `content`. Both options are illustrated below:

```yaml
types:
  Org:
    type: object
    properties:
      name: string
      address?: string
    examples: # multiple examples using an array of different contents
      - content:
          name: Acme
      - content:
          name: Software Corp
          address: 35 Central Street
```

```yaml
types:
  Org:
    type: object
    properties:
      name: string
      address?: string
    examples: # multiple examples identified by their unique key names
      acme:
        content:
          name: Acme
      softwareCorp:
        content:
          name: Software Corp
          address: 35 Central Street
```

The first example using an array of `-content` keys is not valid according to YAML and should not be supported in the spec IMHO. The second uses unique keys which is fine, but we should remove `content` as it is not coherent to how you define a single example like the following:

```yaml
types:
  Org:
    type: object
    properties:
      name: string
      address?: string
    example: # single example
      name: Software Corp
      address: 35 Central Street
```

My suggestion would be to use the same set of properties available in a single example also in examples except that each example needs to be identified by a unique key. The following is a modified version illustrates that:

```yaml
types:
  Org:
    type: object
    properties:
      name: string
      address?: string
    examples: # multiple examples identified by their unique key names
      - acme:
          name: Acme
      - softwareCorp:
          name: Software Corp
          address: 35 Central Street
```

I am looking forward to hear some feedback.

freddrake:
On Fri, Feb 5, 2016 at 6:37 PM, Christian Vogel
<notifications@github.com> wrote:
> My suggestion would be to use the same set of properties
> available in a single example also in examples except that
> each example needs to be identified by a unique key.

I definitely like your example better.


  -Fred

-- 
Fred L. Drake, Jr.    <fred at fdrake.net>
"A storm broke loose in my mind."  --Albert Einstein

petrochenko-pavel-a:
It is definitely more compact and I like it, but I have couple of questions:

1) how any body can annotation an example
2) how any body can give a human readable description to it.

sichvoge:
1) annotating should be not too difficult

```yaml
types:
  Org:
    type: object
    properties:
      name: string
      address?: string
    examples: # multiple examples identified by their unique key names
      - acme:
          (annotation): true
          name: Acme
      - softwareCorp:
          name: Software Corp
          address: 35 Central Street
```

[...]

PR #295, opened 2016-03-11
changes to how to define examples in RAML

sichvoge:
According to #239 - examples should not be represented as a sequence anymore and additionally the syntax of examples changes as well.

When we do the second round of clean-up; the chapter will look a bit different. This PR puts changes for the first iteration into the spec.


usarid:
The examples (no pun intended) should include one or two higher levels i.e. they should have the keywords `example` and `examples` in them. Maybe just put a more fully-worked-out example, with an actual method payload, and some query parameters?

PR #296, opened 2016-03-11
change 'uses' to act like 'requires' 

### LINKED ISSUE #226: Change `uses` to act like "requires" in other languages
The proposal is to change the semantics of `uses` to indicate that the remote library to be used is _available_ as a local asset, using a local namespace. 

Using `uses` does NOT automatically import the remote library's assets into the local file, but it allows the local file to import those assets by referring to them in its contents: e.g., a RAML type fragment file that `uses` a library of remote types can then import one of those types by referring to it, and that remote type will be included as if it were defined locally within the RAML type fragment file.

The namespaces defined in a `uses` statement in a file are only consumable within that file, i.e. they are only used to disambiguate the included libraries from each other within that file.

This proposal also expands the use of `uses` to allow its use in a root ("master") RAML file as well as in all RAML fragment files.

The `uses` keyword may ONLY be used at the root of a RAML file (either a root ["master"] RAML file or a RAML fragment file), and its value MUST be the location of a RAML library file.

This proposal is aimed to simplify the conceptual model of consuming libraries, and better support modularity of API definitions, e.g. by eliminating artificial dependency cycles (https://github.com/mulesoft/api-workbench/issues/29).

sichvoge:
closes #226

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -3075,8 +3075,14 @@ resourceTypes:
   file:
     get:
       is: drm
+      responses:
+        201:
+          body:
+            application/json:
+              type: file-type.File
     put:
       is: drm
+
 ```
 
 ```yaml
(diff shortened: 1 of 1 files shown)


usarid:
The text looks right but the examples look wrong, specifically they don't actually use the namespaced assets.

PR #297, opened 2016-03-12
Fix `describedBy` anchor

nogates:
Fix documentation anchor. Seems that github sets all the anchors in downcase.

Also, Branch `raml-10` has the same mistake.


sichvoge:
@nogates thanks for the PR - will merge it. Regarding the `raml-10` branch; this is one that we will update as soon as `raml-10-rc2` is finished. All changes should go in the latter.

Issue #292, opened 2016-03-10
Merging process of optional query parameter in traits is counter-intuitive


ddenisenko:
Due to the following wordings in the spec, located in Optional Properties subsection of Resource Types and Traits section:
>To accommodate this need, a resource type or trait definition MAY append a question mark ("?") suffix to the name of any object-valued property that should not be applied if it doesn't already exist in the resource or method at the corresponding level. A property is considered object-valued if its value is of type object or of a type that has object at the base of its inheritance hierarchy. The question mark suffix indicates that the value of the property in the resource type or trait should be applied if the property name itself (without the question mark) is already defined (whether explicitly or implicitly) at the corresponding level in that resource or method.

optional query parameter defined in a trait is not merged into a method unless such parameter is already defined in the method itself.

This is not something users expect.

ddenisenko:
Here's the initial input from Uri:

we should clarify: optional properties are supposed to be for methods, and then at some point we thought: hey, why not say it's for everything? But that's before we have ? for anything else and before we had data types. So maybe we should restrict it back down to methods again.

ddenisenko:
Well, the current version contains exactly the quoted wordings. So if we agree that optional query parameters should be merged, that wordings should be restricted (either to methods only, or may be someone can come up with a broader restriction still covering query parameters).

sichvoge:
If we restrict it to methods, optional properties will only be available for resource types. Do we have any use cases where we would need these mechanism in traits. I haven't seen any so far, but maybe someone else knows.

usarid:
You could probably make a case for the items under `responses` (i.e. *if* a method defines a `201` *then* it should behave as follows) and for the items under `securedBy` (i.e. *if* you're secured by OAuth *then* use the following), but for simplicity I would restrict it to just methods, which means as you say that it only applies to resource types.

sichvoge:
OK - will apply that for RC2. Thanks all!

PR #299, opened 2016-03-14
restrict optional properties to methods only 

sichvoge:
close issue #292

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -2065,11 +2065,11 @@ traits:
 
 Parameters may not be used within !include tags, that is, within the location of the file to be included.
 
-### Optional Properties
+### Declaring HTTP Methods as Optional
 
-When defining resource types and traits, it can be useful to capture patterns that manifest several levels below the inheriting resource or method, without mandating the creation of the intermediate levels. For example, a resource type declaration may describe a body parameter that would be used if the API defines a post method for that resource, but applying the resource type to a resource without a post method should not create the post method.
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -2096,29 +2096,7 @@ resourceTypes:
...
-  post:
-    is: [ hasPayload ]
-    # will have application/json body with 'name' required property,
-    # 'address' optional property, but without 'location' property
-```
-
-The POST:/servers method obtains an application/json body with with 'name' required property, 'address' optional property, but without 'location' property.
-
-It is important to note that this feature applies only to object-valued properties; the appending of the optional marker ("?") to a scalar- or array-valued property such as description or security schemes MUST be rejected by RAML parsers.
+Note that a question mark can also appear as the last character of a RAML Type property name. If you wish to use this syntax within a resource type you should escape it with a backslash ( "\?").
```
usarid: I'm quite confused by this: what kind of escape mechanism is this, i.e. at what level is it escaping the question mark -- at the YAML level, or somehow in RAML itself?

usarid:
LGTM other than the one comment that we need to resolve.

versions/raml-10/raml-10.md
sichvoge: I don't think that we actually need that anymore since we restricted optional properties to be applied to methods only. It has been there before bc we defined an optional property related to a trait or resource type to be any object-valued node.

usarid:
So regardless of optional properties, how do you in fact define a data type which has a property whose name happens to end with a question mark?

I've filed a separate issue for that, #317, so please remove that part of the change from this PR and make a change separately that I'll then review.

PR #301, opened 2016-03-14
clarifies that any trait that is applied to a method, gets applied in a specific order

### LINKED ISSUE #300: clarify that traits are applied from left to right
In section "Applying Resource Types and Traits" we need to make sure that we mention how traits defined in a sequence are getting applied. 

As in https://github.com/raml-org/raml-spec/issues/184#issuecomment-189975866 mentioned; they are getting applied from left to right.

sichvoge:
closes #300

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1980,6 +1980,17 @@ traits:
    - rateLimited: !include traits/rate-limited.raml
 ```
 
+Neither resource types nor traits are allowed to incorporate nested resources. Therefore, the following example is invalid:
+
+```yaml
+resourceTypes:
+  hasGroups:
+    get:
+    patch:
+    delete:
+    /groups:
+```
+
 ### Usage
 
 The OPTIONAL **usage** property of a resource type or trait provides instructions on how and when the resource type or trait should be used. Documentation generators MUST convey this property as characteristics of the resource and method, respectively. However, the resources and methods MUST NOT inherit the usage property: neither resources nor methods allow a property named usage.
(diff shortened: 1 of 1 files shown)


usarid:
No need to say that trait definitions cannot incorporate nested resources -- not only does it seem like a random comment, this in any case isn't the part of the spec where we define how traits are defined, so let's keep this section clean by not dragging a bit of the trait-definition section into it.

Also, if we're going to mention anonymous trait definitions, better to mention first the more common usage of an `!include` to pull in the trait, and only mention inline definitions as a last resort, included for consistency with how an `!include` works.

sichvoge:
To do that; I think we need to clarify https://github.com/raml-org/raml-spec/issues/302 first.

usarid:
OK, I clarified #302 .

sichvoge:
Thanks Uri. I added a comment just to be sure and highlight the different notations.

sichvoge:
Also, why do you think that `Trait definitions MUST NOT incorporate nested resources.` should not be mentioned here? The section talks about applying traits and resource types, and this is a restriction for both that a trait does not get applied automatically to any nested resources. Am I wrong?

usarid:
These are two different things: you can clarify that traits don't get *applied* to nested resources, and that's separate from a comment in the trait and resource type definitions section that traits and resource types must not incorporate (have in them) nested resources. You __cannot__ do:
```raml
resourceTypes:
  hasGroups:
    get:
    patch:
    delete:
    /groups:
```

sichvoge:
AH - haha, my bad! Got it.

usarid:
Also, I'm also not seeing this in your PR: "a comment just to be sure and highlight the different notations"

sichvoge:
It's in #302

sichvoge:
I apply the requested changes and we will do another PR if necessary for #302

Issue #306, opened 2016-03-14
Remove sequences and composition in RAML 1.0 for global definitions like traits and resource types


sichvoge:
RAML 1.0 is the first major release, and we want to be very close to one of our missions to introduce simple ways to share and collaborate pieces of your API design. In the past, we introduced concepts to define reusable definitions globally or in separated files, which could be easily included on or composed with other definitions. An example of such may be the following:

**traits.raml**

```yaml
# external traits definition
pagination:
  queryParameters:
    offset:
      description: Skip over a number of elements by specifying an offset value for the query
      type: integer
      example: 20
      default: 0
    limit:
      description: Limit the number of elements on the response
      type: integer
      example: 80
secondTrait:
   description: just a second trait
```

**api.raml**

```yaml
#%RAML 0.8
title: Pagination API

traits: 
  - !include traits.raml # composing with includes from traits.raml
  - otherTrait: # explicit trait definition
      description: just another trait
```

RAML 1.0 takes the concepts of reusability or modularization one step further with introducing fragments and libraries. Both add significant value to the core concepts already mentioned and make some of the previous mechanism introduced above obsolete; especially composition. The above example can be also perfectly expressed using libraries. 

**traits.raml**

```yaml
#%RAML 1.0 Library

# external library

traits:
  pagination:
    queryParameters:
      offset:
        description: Skip over a number of elements by specifying an offset value for the query
        type: integer
        example: 20
        default: 0
      limit:
        description: Limit the number of elements on the response
        type: integer
        example: 80
  secondTrait:
     description: just a second trait
```

**api.raml**

```yaml
#%RAML 1.0
title: Pagination API

uses:
  externalTraits: traits.raml
traits:
  otherTrait:
    description: just another trait
```

Using libraries, and also fragments, are coming with a couple of advantages; one being able to give better support for design-time validation and suggestions, and more. 

Since we believe that these new mechanism will give significantly more benefits to anyone in the RAML community, we effectively decided to remove the ability to compose any external definitions with global definitions in RAML 1.0 RC2. So in any future version of RAML, you will be not allowed to do the following anymore:

```yaml
traits: 
  - !include traits.raml # composing with includes from traits.raml
  - otherTrait: # explicit trait definition
      description: just another trait
```

[...]

PR #320, opened 2016-03-17
removed occurrences of sequences from global definitions

sichvoge:
As per #306, this PR removes the notion of sequences and composition from the RAML 1.0 RC2 spec. There wasn't really anything to do for removing composition as it wasn't mentioned anywhere anyways. All changes has been applied to the existing examples. I also corrected some indention problems.

@usarid not sure if there is something to review here since the only changes were on the examples to remove the support for sequences. The text itself did not explicitly mention what the value of resource types, traits, and security schemes is. Of course that will be added with another iteration as you know. Assigning you nevertheless as it is a `major` change and I want your final confirmation.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -107,12 +107,12 @@ version: v3
 baseUri: https://api.github.com
 mediaType:  application/json
 securitySchemes:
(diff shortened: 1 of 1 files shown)


usarid:
LGTM. All examples should be run through the latest parser once that'd updated so we'll know if they're right.

PR #328, opened 2016-03-21
removed inline + include support 

### LINKED ISSUE #321: Remove support for any inline definition (implicit or explicit) for traits, resource types, and security schemes
Original Discussion: #302 

In RAML  0.8 and 1.0 RC1 you were able to define inline (anonymous) traits, resource types, and security schemes either explicitly or implicitly using `!include`. This syntax is a shortcut that promotes a bad design practice, as you will not be able to reuse the trait in your api unless you include the same file more than once which does not look very DRY. Also the feature is crippled as you can inline only certain types of traits or resource types, those that do not use parameters.

As there might be some advantages, they are very small compared with the complexity and bad design you introduce into the spec and your API design. Therefore, the suggestion is to remove the ability to define inline definitions (implicit or explicit) for traits, resource types, and security schemes. 

So the following are examples that are not valid in RAML 1.0 RC2 anymore:

```yaml
/resource:
  get:
    is:
      - !include trait1.raml
      - !include trait2.raml
```

```yaml
/users:
  type: !include types/collectionBase.raml
```

```yaml
/resource:
  get:
    is:
      - queryParameters:
          query1:
      - queryParameters:
          query2:
```

If one has, for example, a resource type defined in an external fragment file; it is recommended to use the global definition to include it into your RAML definition.

```yaml
resourceTypes:
  baseCollection: !include types/collectionBase.raml
/users:
  type: baseCollection
```

This change is restricted only to traits, resource types, and security schemes; and you will still be able to use that mechanism for any other such as `type`.

```yaml
/users:
  get:
    responses:
      200:
        body:
          type: !include users.json
```

sichvoge:
according to issue #321; this PR removes inline + include support for traits and resource types.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1986,23 +1986,23 @@ A resource type is a partial resource definition that, like a resource, can spec
 
 A trait is a partial method definition that, like a method, can provide method-level properties such as description, headers, query string parameters, and responses. Methods that use one or more traits inherit those traits' properties. Resources and resource types can also use, and thus inherit from, one or more traits, which then apply to all of their methods. Traits are related to methods through a mixing pattern.
 
-Resources may specify the resource type from which they inherit using the type property. The resource type may be defined inline as the value of the type property (directly or via an !include), or the value of the type property may be the name of a resource type defined within the root-level resourceTypes property or in a library.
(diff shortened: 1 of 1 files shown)


usarid:
LGTM

Issue #307, opened 2016-03-14
Remove `discriminator` from any union type definition


sichvoge:
In RAML 1.0 RC1 you are able to define a discriminator for union types. This could introduce a lot of ambiguity when, for example, the types itself have their definition of a discriminator.

To avoid any ambiguity, the suggestion is to remove `discriminator ` from any union type.

```yaml
PersonOrDog: 
   type: Person | Dog
   discriminator: hasTail # will be removed in RC2
```

sichvoge:
Not done! It is still there and was missing. Reopen it until fixed.

PR #333, opened 2016-03-22
removed polymorphism section and moved examples for discriminators up…

### LINKED ISSUE #309: Clarify that 'discriminator' and 'discriminatorValue' can only be defined for global type definitions
Due to its nature, RC2 should explicitly state that defining `discriminator` and `discriminatorValue` should only be available for global type definitions and not for any types that are declared inline.

```yaml
# valid
types:
   Device:
     type: Phone | Notebook
     discriminator: kind
```

```yaml
# invalid
application/json: 
   type: Phone | Notebook
   discriminator: kind
```

### LINKED ISSUE #330: Remove Polymorphism section in favour for a better description for `discriminator` and `discriminatorValue`
The current RC1 section about [Runtime Polymorphism (Discriminators)](https://github.com/raml-org/raml-spec/blob/raml-10/versions/raml-10/raml-10.md#runtime-polymorphism-discriminators) is actually not needed anymore in RC2. 

For RC2, it's enough to add a bit more description for `discriminator` and `discriminatorValue`. 

------------------------------

| Name | Type | Description | 
|:---------|:------:|:-----------------|
| discriminator | `string` | The value must correspond to a name of one of the properties that are defined inside a type declaration. The `discriminator` facet can not be defined for inline type declarations and should only be used for scalar-valued properties.
| discriminatorValue | `string`  | You may use `discriminatorValue` facet if a type also defined the `discriminator` facet. The value of this facet should be a valid value of the property with the name equal to the value of `discriminator` facet; and its being used to identify the declaring type. This value should be unique in the hierarchy of the type.<br/><br/>**Default:** the name of the type where the discriminator is applied to

sichvoge:
… to the object type section; as well as changed description for discriminator and discriminatorValue to close #330

This PR automatically closes #309 and #307 with the new introduced "Discriminator" chapter.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -733,10 +733,10 @@ data:
     kind: employee
 ```
 
-Neither `discriminator` nor `discriminatorValue` are allowed to be defined for any inline type declarations.
+Neither `discriminator` nor `discriminatorValue` are allowed to be defined for any inline type declarations or union types.
 
 ```yaml
-# valid
+# valid when ever there is a key name that can identify a type
 types:
   Device:
     discriminator: kind
@@ -745,11 +745,18 @@ types:
 ```
 
 ```yaml
-# invalid
+# invalid in any inline type declarations
 application/json:
    discriminator: kind
    properties:
      kind: string
+```
+
+```yaml
+# invalid for union types
+PersonOrDog:
+   type: Person | Dog
+   discriminator: hasTail
 ```
 
 ### Array Types
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -461,7 +461,7 @@ All types that have the built-in object type at the root of their inheritance tr
 | additionalProperties? | JSON schema style syntax for declaring maps. See section [Map Types](#map-types) for more information.
 | patternProperties? | JSON schema style syntax for declaring key restricted maps. See section [Map Types](#map-types) for more information.
 | discriminator? | You may use the `discriminator` property to be able to discriminate the concrete type of an individual object at runtime when, for example, payloads contain ambiguous types (achieved via unions or inheritance). The value must correspond to a name of one of the `properties` that are defined inside a type declaration. The `discriminator` property can not be defined for inline type declarations and should only be used for scalar-valued properties. See section [Using Discriminator](#using-discriminator) for more information.
-| discriminatorValue? | You may use the `discriminatorValue` property if a type also defined the `discriminator` property. The value of this property should be a valid value of the property with the name equal to the value of the `discriminator` property; and its being used to identify the declaring type. This value should be unique in the hierarchy of the type. See section [Using Discriminator](#using-discriminator) for more information.<br/><br/>**Default:** the name of the type where the discriminator is applied to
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -460,8 +460,8 @@ All types that have the built-in object type at the root of their inheritance tr
 | maxProperties? | The maximum number of properties allowed for instances of this type.
 | additionalProperties? | JSON schema style syntax for declaring maps. See section [Map Types](#map-types) for more information.
 | patternProperties? | JSON schema style syntax for declaring key restricted maps. See section [Map Types](#map-types) for more information.
-| discriminator? | Type property name to be used as discriminator, or boolean
-| discriminatorValue? | The value of discriminator for the type.
+| discriminator? | You may use the `discriminator` property to be able to discriminate the concrete type of an individual object at runtime when, for example, payloads contain ambiguous types (achieved via unions or inheritance). The value must correspond to a name of one of the `properties` that are defined inside a type declaration. The `discriminator` property can not be defined for inline type declarations and should only be used for scalar-valued properties. See an example on how to use `discriminator` [here](#using-discriminator-example).<br/><br/>**Default:** the name of the type
```
petrochenko-pavel-a: If the type is suitable to AC then `discriminator` property is not needed

versions/raml-10/raml-10.md
```diff
@@ -659,6 +659,74 @@ types:
         type: Method
 ```
 
+#### Using Discriminator Example
```
petrochenko-pavel-a: I think we need some more intro wording here.

versions/raml-10/raml-10.md
```diff
@@ -659,6 +659,74 @@ types:
...
+data:
+  - name: A User
+    userId: 111
+    kind: User
+  - name: An Employee
+    employeeId: 222
+    kind: Employee
+```
+
+You can also override the value of the discriminator for each individual concrete class. In the following example we will replace the default type discriminator values by their lowercase versions using `discriminatorValue`:
```
petrochenko-pavel-a: You probably mean `discriminatorValue`

versions/raml-10/raml-10.md
sichvoge: That is why we have the default right? The automatic classification (AC) uses that, am I correct? And I don't think we should directly mention AC in this context. So what do you want me to change?

versions/raml-10/raml-10.md
sichvoge: Don't understand. What are you missing or referring to?

versions/raml-10/raml-10.md
sichvoge: What do you think about? That is not a separated "Discriminator" section and only illustrates an example of using it. Is the description for the specific properties not enough? Are you thinking about just giving an intro to the example?

versions/raml-10/raml-10.md
petrochenko-pavel-a: >And I don't think we should directly mention AC in this context.
Agree. I actually did not like this phrase:
>You may use the `discriminator` property to be able to discriminate the concrete type of an individual object at runtime.

Basically it leads you to opinion that you need to use it in any case. In fact I like description here pretty much:
https://github.com/mulesoft-labs/typesystem-docs/blame/master/typesystem.md#L1197

versions/raml-10/raml-10.md
petrochenko-pavel-a: yep, I do not actually know how to write it correctly. But this looks like a bit unexpected.

versions/raml-10/raml-10.md
petrochenko-pavel-a: >the value of the discriminator for each individual concrete class
In fact you are overriding `discriminatorValue`

sichvoge: Then it was wrong in RC1 as well. Modifying it.

petrochenko-pavel-a:
Looks good to me.

Issue #184, opened 2015-09-30
Override Rules


usarid:
Certain behaviors were not explicit in RAML 0.8; this fixes them. This includes:
* Peer resources that resolve to the same path, e.g. `/users` with a sub-resource `/{id}`, alongside `/users/{id}`
* Peer resources with similar URI parameters, e.g. `/{pictureId}` and `/{videoId}`
* Duplicate methods on a resource
* Multiple occurrences of the same media type in a response
* The order in which to apply multiple traits (and now types, since there is multiple inheritance)

There are a few others to clarify too.

sichvoge:
Can you clarify a little bit more what you want to fix?

sichvoge:
This is related to #199 I guess.

usarid:
Yes, we need to probably bring back the overrides section, per #199 

Specifically to the points I raised in this current issue:
* Peer resources that resolve to the same path, e.g. `/users` with a sub-resource `/{id}`, alongside `/users/{id}` -- to be addressed in the overrides section
* Peer resources with similar URI parameters, e.g. `/{pictureId}` and `/{videoId}` -- ALLOWED, to be addressed in the overrides section
* Duplicate methods on a resource -- can not happen as it would be inconsistent YAML (because of duplication of properties)
* Multiple occurrences of the same media type in a response -- can not happen as it would be inconsistent YAML (because of duplication of properties)
* The order in which to apply multiple traits (and now types, since there is multiple inheritance) -- from left to right, but this too should be described explicitly in the section explaining how to apply traits and resource types, and in the section on multiple extensions (inheritance) of types.

sichvoge:
> Peer resources that resolve to the same path, e.g. /users with a sub-resource /{id}, alongside /users/{id} -- to be addressed in the overrides section

@usarid - what is the default behaviour? I remember discussions about "marking" them as ambiguous and maybe add a warning in IDEs or so.

sichvoge:
And also what about overloading definition that has been declared in traits and resource types?

sichvoge:
Added a PR that introduces the new overloading section. @usarid can you please check and add any comments you might have. I know we discussed briefly that we might also not allow any  overloading of resources, for example:

```yaml
/users:
  /foo:
/users/foo
```

but allow things like:

```yaml
/users/{userId}:
/users/{username}:
/users/me:
```

The chapter at least gives us a base for discussion.

svacas:
[...]

PR #336, opened 2016-03-24
introduces new overloading section

sichvoge:
According to #184 we need to explicitly define how a RAML processor should act in certain cases where one overloads, for example, a resource or method. This section gives an overview about the rules that are mentioned in #184.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1980,6 +1980,133 @@ types:
             error: Amount cannot be negative
 ```
 
+## Overloading Rules
+
+RAML overloading allows to define similar resources, which vary only on "secondary" characteristics, such as URI parameter types and methods.
+
+### Motivation
+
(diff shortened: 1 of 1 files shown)


usarid:
I proposed largely eliminating overloading altogether in #184 so we need to get closure on that before finalizing this PR. How can we get comfort around that?

sichvoge:
Guess we need to answer final questions on the issue first. Do you really want to remove it completely or are we providing at least some kind of information around ambiguity and how to detect them?

usarid:
Once we resolve #184 we'll see whether there's anything left to say in this section, vs just specifying the rule in the section for resources.

sichvoge:
Agree; I don't think that the section is necessary if we only go for the resources. Everything else is not allowed due to the YAML nature and unique keys.

usarid:
OK, so given my last statement about not having overloading, we should not have this section, so I can't accept this PR which introduces this section.

Issue #286, opened 2016-03-09
Map declaration is not yaml 1.2 compliant


robvanmieghem:
Map declaration (https://github.com/raml-org/raml-spec/blob/raml-10/versions/raml-10/raml-10.md#map-types):
```
#%RAML 1.0
title: My API With Types
types:
  MapOfNumbers:
    type: object
    properties:
      []:
        type: number
```

The standard golang yaml parser bails with the following error:

`yaml: invalid map key: []`

The online yaml parser http://yaml-online-parser.appspot.com bails with the following error:
```
ERROR:

while constructing a mapping
found unacceptable key (list objects are unhashable)
```

I suggest to drop this way of declaring maps since the inline map declarations are a lot nicer anyway:
```
#%RAML 1.0
title: My API With Types
types:
  MapOfNumbers:
    type: number{}
```

sichvoge:
I do agree that this semantic shortcut is quite helpful, but we should be consistent with our approach to have an extended version and a shortcut. What do you think about using the following notation:

```yaml
#%RAML 1.0
title: My API With Types
types:
  MapOfNumbers:
    type: object
    properties:
      "[]":
        type: number
```

robvanmieghem:
That's indeed a lot better, this also works with https://github.com/raml-org/raml-js-parser-2

I think it would be good to update the spec and documentation to use `"[]"` instead of `[]`

sichvoge:
@robvanmieghem it's strange that YAML parser is complaining actually. According to the spec any key inside a map can be again a node. So theoretically sequences should be supported. :/

petrochenko-pavel-a:
>I think it would be good to update the spec and documentation to use "[]" instead of []
@robvanmieghem , I agree. Currently we have an  ugly border case in the parser to support []

robvanmieghem:
> it's strange that YAML parser is complaining actually. According to the spec any key inside a map can be again a node. So theoretically sequences should be supported. :/

@sichvoge I don't want to start a discussion on the yaml spec interpretation here but it is a fact that most yaml parsers do not accept this syntax as being valid. For a spec as RAML I think it is not a good idea to use a border case since this will not improve RAML adoption and compatibility.

usarid:
I agree, and I think having `"[]"` is possibly even worse (who's going to remember that?), so perhaps we should find another notation...

sichvoge:
@robvanmieghem any perspective on an alternative notation or do you think that `"[]"` is simple enough?

[...]

PR #337, opened 2016-03-24
changed behaviour of additional properties

sichvoge:
removed map types chapter in favour for a more clear additional properties section that belongs to the object type chapter; and added the changes according to issue #286


petrochenko-pavel-a:
Looks good.

Issue #308, opened 2016-03-14
Clarify that `schema` and `type` are mutually exclusive


sichvoge:
RAML 1.0 RC2 should clearly state that `schema` and `type` are mutually exclusive. Therefore, the following should be invalid in any future version of RAML.

```yaml
/resource:
  get:
    responses: 
      200:
        body: 
          application/json:
            schema: 
            type:
```

You are still allowed to have global definitions for `schemas` or `types`, although it is recommended to use `types` only.

svacas:
What's the suggested approach for users that already have their entities in XML schema and/or JSON schema?

usarid:
They can keep using them and just refer to them with the keyword `type` or `types` (which are not deprecated, vs their synonyms `schema` and `schemas` which _are_ deprecated), or they can see if they can model their data in RAML data types now.

PR #338, opened 2016-03-24
make clear that schema and type are mutual exclusive

sichvoge:
closes issue #308


petrochenko-pavel-a:
Looks good.

usarid:
LGTM, though I would use every opportunity, including this one, to re-emphasize that they are **synonymous**, and that the `schema` synonym is deprecated so in the future only the `type` synonym will be allowed.

Issue #304, opened 2016-03-14
What is a partial method or resource?


sichvoge:
In RAML 0.8 and 1.0 we define a trait or resource type as "partial method/resource", be don't really specify what this "partial" inherits.

My initial concern about that and looking at different implementation is that, for example, some support to define `is` in a trait and some aren't. The spec is not clear about that.

```yaml
traits: 
  test:
    is: [ pageable ]
  pageable:
```

I am fine with either, but we should be more explicit on what "partial" means, or even if we need to mention that it is partial.

sichvoge:
We should also think about if we really want to allow inheritance on trait level since we have composition already.

```yaml
traits: 
  test:
    is: [ pageable ]
  pageable:
```

vs

```yaml
traits: 
  test:
  pageable:
/resource:
  is: [ pageable, test ]
```

freddrake:
I expect inheritance among traits to be very useful; my first attempt to write a RAML spec included an attempt to do just that.  It seems very reasonable to have traits that require another trait to also be used for the affected resource; requiring that to be unpacked for each use doesn't really make sense, and is likely to be a source of bugs in specifications.

svacas:
According to the spec `is` can be used in two places: inside a method and inside a resource. Any other place should be an error. Also inheritance is only allowed for resource types, not traits.

usarid:
It's a mixin type of composition, whether you regard it as inheritance or not. As for whether it's according to the spec: a) RAML 0.8 may have been ambiguous about it, but b) there's no reason we cannot allow it in RAML 1.0 and remove the ambiguity. And I do believe it contributes to the modularity of API specs, as @freddrake commented: just as you can build up resource types from resource types and traits and reuse the composed ones, you should be able to concoct traits from other traits and reuse the composed ones.

Let's clarify that this is explicitly allowed in RAML 1.0.

usarid:
@sichvoge if there are other facets of methods that should be explicitly clarified and included/excluded in the definition of traits, please identify them and reassign to me.

Please also take a look at the facets allowed in the `describedBy` facet of security schemes to see if we need to clarify there too.

PR #342, opened 2016-03-30
removed partial method/resource text 

sichvoge:
In favour for more clarity according to #304. Also moved redundant content around for a more consistent approach.

The sections in the chapter for resource types and traits were not really well ordered. I moved few things and also tried to consolidate some of the redundant content.


usarid:
LGTM

Issue #317, opened 2016-03-16
Mechanism to allow question mark in data type property names


usarid:
When the `required` facet on a property is specified explicitly in a type definition, any question mark in its property name is treated as part of the property name rather than as an indicator that the property is optional.

For example, in
```raml
types:
  profile:
    properties:
      preference?:
        required: true
```
the `profile` type has a property whose name is `preference?` (including the trailing question mark) and that is required. The same property could be made optional in two equivalent ways:
```raml
types:
  profile:
    properties:
      preference?:
        required: false
```
or
```raml
types:
  profile:
    properties:
      preference??:
```

PR #344, opened 2016-03-30
clarified question mark in property names

sichvoge:
according to #317

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -504,10 +504,6 @@ Properties of object types are defined using the OPTIONAL **properties** propert
 
 In addition to the properties available in normal type declarations, properties can specify whether they are required and provide an optional default value.
 
-Note:
-
-When an Object Type does not contain the "properties" property, the object is assumed to be unconstrained. That means, it may contain any properties of any type.
-
 | Property  | Description |
 |:----------|:----------|
 | default? | Provides default value for a property.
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -529,6 +525,53 @@ types:
...
 
+In order to achieve a more "object oriented" experience, a series of shortcuts are available (see [Shortcuts and Syntactic Sugar](#shortcuts-and-syntactic-sugar)). The example below shows a common idiom:
+
+```yaml
+#%RAML 1.0
+title: My API With Types
+types:
+  Person:
+    properties:
+      name: string # equals to `type: string`
```
usarid: Clarify that the *value* is equivalent to `type: string`, or that the *line* `{ name: string }` is equivalent to `{ name: { type: string } }`.

versions/raml-10/raml-10.md
```diff
@@ -504,10 +504,6 @@ Properties of object types are defined using the OPTIONAL **properties** propert
 
 In addition to the properties available in normal type declarations, properties can specify whether they are required and provide an optional default value.
 
-Note:
-
-When an Object Type does not contain the "properties" property, the object is assumed to be unconstrained. That means, it may contain any properties of any type.
-
```
usarid: I'm confused by this: why do we want to lose this statement about the "properties" property?

Also, can't every type allow the `default` facet?

sichvoge: Agree, looking at current RC1 implementation; `default` is available for every type declaration. @petrochenko-pavel-a can you comment.

petrochenko-pavel-a: As for me it is a good to have `default` available for every type definition.

Issue #319, opened 2016-03-16
Simplify inheritance rules to allow more freedom on multiple inheritance of union types etc


sichvoge:
RC1 added a lot of restrictions to how you model, for example, inheritance with the RAML type system. To also allow use cases like multiple inheritance of union types or inherit from an array type, these rules should be change to be more flexible/open. That also introduce simpler rules for any type validation since there are less corner cases. 

Here is a perfect example that illustrates the usefulness of, for example, being able to extend from union types and object types at the same moment:

```yaml
#%RAML 1.0
title: Play with named parameter typing
types:
  lat-long: # lat & long required; mutually exclusive with location
    properties:
      lat: number
      long: number
  loc: # location required; mutually exclusive with lat & long
    properties:
      location:
  paging: # each is optional, not exclusive with anything
    properties:
      start?: number
      page-size?: number
/locations:
  get:
    queryString:
      type: [paging,  lat-long | loc ]
      examples:
        first:
          content:
            start: 2
            lat: 12
            long: 13
        second:
          content:
            start: 2
            page-size: 20
            location: 1,2
        third:  # not valid
          content:
            lat: 12
            location: 2
          strict: false # because it's not valid
```

Or the following should be valid in RC2 too:

```yaml
types:
   Number1:
     type: number
     minimum: 4
   Number2:
     type: number
     maximum: 10
   Number3: [ Number1, Number2]
```

Therefore, the complete set of rules in RC1 can be chimed down to only these few in RC2:

* You cannot create cyclic dependencies when inheriting
* You cannot extend from external types including JSON Schemas and XSDs
* extending a union type (or a type having a union type anywhere in its inheritance chain) is allowed if and only if all combinations of extensions you get by distributing across all the "option" types in the union type are valid

The last rule is very important for the support of union types and inheritance (possible multiple inheritance). So let's clarify that a bit more.

First, any union type is built from a set of parent types which are used to compose it. Lets name these types `option types` in the future. In the following example:

```raml
types:
   DogOrCat: Dog | Cat
```
`DogOrCat` union type is composed from two option types `Dog` , `Cat`

Now lets review a case with multiple inheritance one more time. Imagine we have the following type definition:

```raml
types:
   HomeAnimal: [ HasHome ,  Dog | Cat ]
```

In this case type `HomeAnimal` has two base types `HasHome` and anonymous union type defined by a type expression - `Dog | Cat`  

[...]

PR #345, opened 2016-03-30
simplified inheritance restrictions 

sichvoge:
This PR simplifies the current inheritance restrictions according to #319

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1211,8 +1211,6 @@ types:
     type: [ A, B ]
 ```
 
-Note: Multiple inheritance is only allowed if all Type Expressions are simple object Types. See [Inheritance Restrictions](#inheritance-restrictions).
-
 ##### Property Overrides
 
 If multiple parent types define a property with the same name:
@@ -1279,15 +1277,26 @@ Validation of instances of external types is delegated to standard json/xsd sche
 
 #### Inheritance Restrictions
 
-* You cannot inherit from types of different kind at the same moment ( kinds are: union types, array types, object types, scalar types )
-* You cannot inherit from types extending union types ( ex: you cannot extend from `Pet` if `Pet = Dog | Cat` )
-* You cannot inherit from multiple primitive types ( Multiple inheritance is only allowed for object types )
-* You cannot inherit from a type that extends Array type ( this will result in simple type aliasing/wrapping )
-* Facets are always inherited
-* You can fix a previously defined facet to a value if the facet is defined on a superclass
-* Properties are only allowed on object types
 * You cannot create cyclic dependencies when inheriting
-* You cannot inherit from "external" types
+* You cannot extend from external types including JSON Schemas, XSDs, and DataType fragments
+* When you are extending from multiple types of different kind (except of union types); validation of resulting type is done by applying usual type validation rules
+* extending a union type (or a type having a union type anywhere in its inheritance chain) is allowed if, and only if, all combinations of extensions you get by distributing across all the types that are part in the union type are valid
+
+The last rule is important for the support of union types and inheritance (possible multiple inheritance). For example, imagine the following type declaration:
+
+```yaml
+types:
+   HomeAnimal: [ HasHome ,  Dog | Cat ]
+```
+
+The type `HomeAnimal` has two base types `HasHome` and an anonymous union type defined by a type expression - `Dog | Cat`. Testing if the `HomeAnimal` type is valid involves taking each of its base types, and checking that a type which is a derived type of this type and each of the union types is a valid type. That means a RAML processor must test that `[HasHome, Dog]` and `[HasHome, Cat]` are valid types. When extending from two union types, a RAML processor should do the same for every possible combination. For example, in:
+
+```yaml
+types:
+   HomeAnimal: [ HasHome | IsOnFarm ,  Dog | Cat | Parrot ]
+```
+
+a RAML processor must test for all 6 possible combinations: `[HasHome, Dog ]`, `[HasHome, Cat ]`, `[HasHome, Parrot]`, `[IsOnFarm, Dog ]`, `[IsOnFarm, Cat ]`, and `[IsOnFarm, Parrot]`.
 
 ### Shortcuts and Syntactic Sugar
 


petrochenko-pavel-a:
LGTM

versions/raml-10/raml-10.md
```diff
@@ -1279,15 +1277,26 @@ Validation of instances of external types is delegated to standard json/xsd sche
...
-* You cannot inherit from types extending union types ( ex: you cannot extend from `Pet` if `Pet = Dog | Cat` )
-* You cannot inherit from multiple primitive types ( Multiple inheritance is only allowed for object types )
-* You cannot inherit from a type that extends Array type ( this will result in simple type aliasing/wrapping )
-* Facets are always inherited
-* You can fix a previously defined facet to a value if the facet is defined on a superclass
-* Properties are only allowed on object types
 * You cannot create cyclic dependencies when inheriting
-* You cannot inherit from "external" types
+* You cannot extend from external types including JSON Schemas, XSDs, and DataType fragments
+* When you are extending from multiple types of different kind (except of union types); validation of resulting type is done by applying usual type validation rules
```
usarid: This is cryptic: if it's just "usual validation rules" then why do we call it out here? And what actually does "usual validation rules" mean?

versions/raml-10/raml-10.md
```diff
@@ -1279,15 +1277,26 @@ Validation of instances of external types is delegated to standard json/xsd sche
...
-* You cannot inherit from types of different kind at the same moment ( kinds are: union types, array types, object types, scalar types )
-* You cannot inherit from types extending union types ( ex: you cannot extend from `Pet` if `Pet = Dog | Cat` )
-* You cannot inherit from multiple primitive types ( Multiple inheritance is only allowed for object types )
-* You cannot inherit from a type that extends Array type ( this will result in simple type aliasing/wrapping )
-* Facets are always inherited
-* You can fix a previously defined facet to a value if the facet is defined on a superclass
-* Properties are only allowed on object types
 * You cannot create cyclic dependencies when inheriting
-* You cannot inherit from "external" types
+* You cannot extend from external types including JSON Schemas, XSDs, and DataType fragments
```
usarid: You cannot extend RAML data type fragments? Are they not just data types?

usarid:
See my 2 comments, otherwise things look OK to me.

versions/raml-10/raml-10.md
sichvoge: @petrochenko-pavel-a can correct me here, but that is related to the fact that you can't extend using `type: !include ...` due to the nature of the `!include` tag. You need to give the datatype fragment an identifier first before using it. For example,

```yaml
types:
  External: !include external-dt.raml
  Internal:
    type: External
```

But that would mean you are not allowed to extend from a DataType in another DataType. I mentioned that before being a limitation, but only recall that we are not going for this. The only possibility I see here is to use the same notation that we use from the `extend` node, or in `uses` where we are not using `!include` anymore. So you could do something like that instead:

```yaml
type: external-dt.raml
```

The downside is that we would more and more go away from needing `!include` imho.

petrochenko-pavel-a:
>@petrochenko-pavel-a can correct me here, but that is related to the fact that you can't extend using type: !include ... due to the nature of the !include tag. 

Exactly.

>This is cryptic: if it's just "usual validation rules" then why do we call it out here? And what actually does >"usual validation rules" mean?
okey, it is probably better to say - as usual.  rules: 

Basic principles

In addition to the structural validation of the type definitions, we use the following principles to determine if certain type definition is semantically valid or not.

1) There is no sense in describing the types having no valid instances. So if RAML processor can prove that it is impossible to describe a JSON structure which will be a valid instance of the type then RAML processor should flag this type definition as invalid

2) If you are using your data type to represent XML then, RAML processor should ensure that transformation of data type instance to XML is isomorphic.

3) Restrictions on custom facets should be full filled (All required facets should be filled and filled with a correct values)

4) Pattern property names should not conflict with normal property definitions (because it leads to inconsistency when transforming to JSON schema)

5) You can not override required property to be optional.

6) you can not declare facets with the names of built-in facets or declared facets available in the hierarchy of the current type

7) facet names can not start with '('

8) you can not declare types with names equal to the names of built-in types.

9) all facets declared on the types should be valid according to their validation rules.

10) you can not use external types anywhere in type expressions (you can not extend from them or use them in facets declarations)

versions/raml-10/raml-10.md
sichvoge: Pavel commented below, and I think we can add these rules to the new chapter that will be introduced with my PR https://github.com/raml-org/raml-spec/pull/351.

versions/raml-10/raml-10.md
usarid: I don't want to add new capabilities, I just think the statement that we cannot extend from external data type fragments is confusing. The way to consume a data type fragment is to `!include` it, and since we don't allow the value of `type` to be an inline type definition (you cannot have `type: { type: ... }`, it's clear that syntactically you cannot run into this situation. But the types defined through data type fragments may be extended, of course, whereas XML and JSON schemas may not -- that's why I found it confusing. I think you can just change "You cannot extend from external types including JSON Schemas, XSDs, and DataType fragments" to "You cannot extend from external types including JSON schemas and XML schemas" and be done.

usarid:
Do these validation rules cited by @petrochenko-pavel-a appear somewhere already, or do they need to be added to the spec? I'm no longer sure what is the appropriate PR here.

versions/raml-10/raml-10.md
sichvoge: I still think we should explicitly call out that you can't do `type: !include ...` since I know that this was not clear in RC1 and has been tried numerous times already by other people.

sichvoge:
They are not explicitly in a specific chapter. Some of them has been mentioned explicitly or implicitly in their respective chapters. For example 7) will be defined in the section for user-defined facets. We could add an additional chapter that highlights or gives an overview about a set of rules that each RAML processor needs to conform when validating types; and then refer back to that chapter in https://github.com/raml-org/raml-spec/pull/345#discussion_r58003348. What do you think?

petrochenko-pavel-a:
As for me having a clear chapter for this is a good idea.

svacas:
```
2) If you are using your data type to represent XML then, RAML processor should ensure that transformation of data type instance to XML is isomorphic.
```
a couple of questions about that statement:

- Is there a list of features of the type system that are rendered invalid when used in an XML media type context? Pattern properties is one that comes to my mind.
- Where can I find the rules for RAML Type -> XSD and RAML Type -> JSON Schema mapping?

petrochenko-pavel-a:
>Where can I find the rules for RAML Type -> XSD and RAML Type -> JSON Schema mapping?
They are not part of the spec so not formalized at this moment. But you can look in early documents 

https://docs.google.com/document/d/1ttuiO7lXgZSPABt3v5N70wpjADL12Vozh-g7FMqE9tY/edit
https://docs.google.com/document/d/1fVjh8VzSqjCH7o8YbMdsInxhmTeKIuNjlk5XscUU7Kw/edit
(they were written pretty long time ago so partially obsolete)

and in Java typesystem implementation playground.

https://github.com/mulesoft-labs/raml-typesystem-java

Regards,
Pavel

sichvoge:
Closing that as invalid. This PR has been covered by couple of smaller PRs to remove inconsistency and repetitive information.

Issue #347, opened 2016-03-31
Facets are part of any type (object, array, or scalar)


sichvoge:
Currently the RC1 spec has an additional property called `facets` that adds user defined restrictions to scalar types. Actually these restrictions can be applied to any other type (object, scalar, or array) except file. That needs to be  more clear in RC2.

Issue #311, opened 2016-03-14
Clarify any user-defined facets related rules


sichvoge:
Currently, the specification is not clear if there are any rules to user-defined facets that avoid ambiguity or conflicts with existing RAML features. Therefore, we should add the following clarifications:

a) you can not start facet name with `(` (this is needed to avoid ambiguity with annotations).
b) you can not redeclare built-in facets for example you can not declare facet with name `properties` in the type which inherits from object type.
c) you can not redeclare facets which were defined earlier in hierarchy any more. this also means that you can not inherit facets with same name from two different types (done to make spec rules a bit simpler)

PR #348, opened 2016-03-31
clarified rules for user defined facets and also made sure that the facets property can be applied to any type except file.

sichvoge:
related to #311 and #347

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -440,7 +440,7 @@ declarations may have; certain type declarations may have other properties.
 | displayName? | An alternate, human-friendly name for the type
 | description? | A longer, human-friendly description of the type
 | (&lt;annotationName&gt;)? | Annotations to be applied to this type. Annotations are any property whose key begins with "(" and ends with ")" and whose name (the part between the beginning and ending parentheses) is a declared annotation name. See section on [Annotations](#annotations) for more information.
-| facets? | A map of user defined restrictions that are being inherited by a specific sub type. See section [User defined Facets](#user-defined-facets) for more information.
+| facets? | A map of additional, user-defined restrictions that will be inherited and applied by any extending subtype. See section [User defined Facets](#user-defined-facets) for more information.
 | xml? | The ability to configure serialization of an instance of this type into XML. See section [XML Serialization of Type Instances](#xml-serialization-of-type-instances) for more information.
 
 The `schema` and `type` properties are mutually exclusive and synonymous: processors MUST NOT allow both to be specified (explicitly or implicitly) inside the same type declaration. Therefore, the following examples are both invalid.
@@ -800,11 +800,11 @@ types:
 
 ### Scalar Types
 
-RAML defines a set of built-in scalar types and each of them has a predefined set of restrictions. All these types, except the file type, may have an additional `enum` property.
+RAML defines a set of built-in scalar types and each of them has a predefined set of restrictions. All these types, except the file type, may have an additional `enum` facet.
 
-|Property Name | Description |
+| Name | Description |
 |:--------|:------------|
-| enum? | Enumeration of possible values for this built-in scalar type. The value is an array containing string representations of possible values, or a single string if there is only one possible value.
+| enum? | Enumeration of possible values for this built-in scalar type. The value is an array containing representations of possible values, or a single value if there is only one possible value.
 
 Example usage of enums:
 
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -956,6 +956,8 @@ A RAML processor must follow specific rules when validating user-defined facets:
 * You can not redeclare built-in facets; for example you can not declare facet with name `properties` in any type which inherits from object type.
 * You can not redeclare facets which has been defined in a super type. This also means that you can not inherit facets with the same name from two different types.
 * You can not validate an instance of a type against user-defined facets.
+* A RAML processors MAY be not required to validate an instance of a type against user-defined facets.
+* A sub-type does not automatically inherit the user-defined facets from its parent type; hence a type that extends from `PossibleMeetingDate` must not define any of the user-defined facets that are part of `CustomDate`.
 
 ### Determine Default Types
 
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -778,11 +778,11 @@ types:
 
 ### Scalar Types
 
-RAML defines a set of built-in types and each of them has a predefined set of restrictions. All built-in types, except the file type, can declare an `enum` property.
+RAML defines a set of built-in scalar types and each of them has a predefined set of restrictions. All these types, except the file type, can have an additional `enum` property.
 
 |Property Name | Description |
 |:--------|:------------|
-| enum? | Enumeration of possible values for this built-in type. The value is an array containing string representations of possible values, or a single string if there is only one possible value.
+| enum? | Enumeration of possible values for this built-in scalar type. The value is an array containing string representations of possible values, or a single string if there is only one possible value.
 
 Example usage of enums:
 
@@ -910,9 +910,9 @@ The ​**file**​ type can be used to constrain the content to send through for
 | minLength? | Specifies the parameter value's minimum number of bytes. Value MUST be equal or greater than 0. Defaults to 0
 | maxLength? | Specifies the parameter value's maximum number of bytes. Value MUST be equal or greater than 0. Defaults to 2147483647
 
-### User defined Facets
+### User-defined Facets
 
-RAML provides a way to also declare user defined restrictions (facets) that extend the predefined set of built-in restrictions for any scalar, array or object type. These can be defined using the `facets` property and its value is a map of user defined facets, specifically the value follows the syntax of a [property declaration](#property-declarations), that may or must be defined in any sub type that extend from the type that declares these facets. For example:
+RAML provides a way to also declare user-defined restrictions (facets) that extend the predefined set of built-in restrictions for any scalar, array or object type. These can be defined using the `facets` property and its value is a map of user defined facets, specifically the value follows the syntax of a [property declaration](#property-declarations). Depending on whether a facet is optional or required, any subtype may or must define the facets that are declared in the type it extends. For example:
 
 ```yaml
 #%RAML 1.0
@@ -928,11 +928,12 @@ types:
     noHolidays: true
 ```
 
-A RAML processor must follow specific rules when validating user defined facets:
+A RAML processor must follow specific rules when validating user-defined facets:
 
 * You can not start facet name with `(` (this is needed to avoid ambiguity with annotations).
 * You can not redeclare built-in facets; for example you can not declare facet with name `properties` in any type which inherits from object type.
 * You can not redeclare facets which were defined earlier in hierarchy any more. This also means that you can not inherit facets with same name from two different types.
+* You can not validate an instance of a type against user-defined facets.
 
 ### Type Expressions
 
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -962,13 +962,6 @@ In the above example, the possibility of restricting date instances based on whe
 
 Because user-defined facets are by definition not built into this RAML specification, and hence their semantic may not be understood by all RAML processors, a RAML processor may or may not choose to use user-defined facets on a type in validating instances of that type.  In the example above, a RAML processor may or may not assign a meaning to `noHolidays` and so may choose to ignore the `noHolidays: true` value in validating instances of `PossibleMeetingDate`.
 
-A RAML processor must follow specific rules when validating user-defined facets:
-
-* You can not start facet name with `(` (this is needed to avoid ambiguity with annotations).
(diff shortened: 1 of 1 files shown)


petrochenko-pavel-a:
LGTM

versions/raml-10/raml-10.md
```diff
@@ -777,18 +778,28 @@ types:
 
 ### Scalar Types
 
-RAML provides a predefined set of Scalar Types. You can "extend" these types and add further restrictions. These restrictions are called "facets" and they exist both for scalar and object types. Each scalar type has a predefined set of facets.
+RAML defines a set of built-in types and each of them has a predefined set of restrictions. All built-in types, except the file type, can declare an `enum` property.
```
usarid: Aren't `object` and `array` built into RAML? In that case, "All built-in types, except the file type, can declare an `enum` property" is not true.

versions/raml-10/raml-10.md
```diff
@@ -899,55 +910,29 @@ The ​**file**​ type can be used to constrain the content to send through for
...
-      format: string
+A RAML processor must follow specific rules when validating user defined facets:
 
-  MyDate:
-    type: Date
-    format: 'dd.mm.yyyy'
-```
+* You can not start facet name with `(` (this is needed to avoid ambiguity with annotations).
+* You can not redeclare built-in facets; for example you can not declare facet with name `properties` in any type which inherits from object type.
+* You can not redeclare facets which were defined earlier in hierarchy any more. This also means that you can not inherit facets with same name from two different types.
```
usarid: What does "earlier in hierarchy any more" mean?

versions/raml-10/raml-10.md
```diff
@@ -899,55 +910,29 @@ The ​**file**​ type can be used to constrain the content to send through for
 | minLength? | Specifies the parameter value's minimum number of bytes. Value MUST be equal or greater than 0. Defaults to 0
 | maxLength? | Specifies the parameter value's maximum number of bytes. Value MUST be equal or greater than 0. Defaults to 2147483647
 
-#### Enums
-
-All custom primitive types, except for the file type, can declare an "enum" property. It should contain a set of fixed values.
+### User defined Facets
 
-Example usage of enums
+RAML provides a way to also declare user defined restrictions (facets) that extend the predefined set of built-in restrictions for any scalar, array or object type. These can be defined using the `facets` property and its value is a map of user defined facets, specifically the value follows the syntax of a [property declaration](#property-declarations), that may or must be defined in any sub type that extend from the type that declares these facets. For example:
```
usarid: Specify that "may or must" corresponds to whether that facet is optional or required, respectively, in the extending type.

usarid:
I don't think this is quite done. We need to say that processors are not required to validate instances against user-defined facets because they may not understand their meaning, so if such validation occurs it is beyond the purview of RAML. In your example, in general a RAML processor will not know how to validate dates against `noHolidays`.

User-defined facets are similar to annotations; the value they add occurs when they are extended, as they can force the extending types to specify their values, while annotations cannot be forced on anything.

versions/raml-10/raml-10.md
sichvoge: Indeed, the missing piece here is `built-in scalar types`. Will change that.

versions/raml-10/raml-10.md
petrochenko-pavel-a: This means in super types.

sichvoge:
Made adjustment based on the feedback. See last commit.

versions/raml-10/raml-10.md
```diff
@@ -921,55 +932,30 @@ The ​**file**​ type can be used to constrain the content to send through for
...
+A RAML processor must follow specific rules when validating user-defined facets:
 
-  MyDate:
-    type: Date
-    format: 'dd.mm.yyyy'
-```
+* You can not start facet name with `(` (this is needed to avoid ambiguity with annotations).
+* You can not redeclare built-in facets; for example you can not declare facet with name `properties` in any type which inherits from object type.
+* You can not redeclare facets which has been defined in a super type. This also means that you can not inherit facets with the same name from two different types.
+* You can not validate an instance of a type against user-defined facets.
```
usarid: >You can not validate an instance of a type against user-defined facets.

Sure you can! That's the whole reason to have user-defined facets. It's just that processors are not required to validate against them, because they may not understand that user-defined facet. But if they do understand them, then they may make use of them, in particular using them for validation.

usarid:
I added one comment, but now I realize we have to do more here. In your example, the `CustomDate` type defines 2 facets, which the extending type then needs to specify, but does the extending type define those facets too -- so something extending `PossibleMeetingDate` also needs to specify those facets? And can `CustomDate` itself specify those facets, or are they only available to extending types? @petrochenko-pavel-a please help us here a bit.

petrochenko-pavel-a:
>but does the extending type define those facets too
no facets are fixed after they are specified.

>And can CustomDate itself specify those facets
No it can not redeclare existing facets, but it can define new ones if it wants.

sichvoge:
Based on Pavel's feedback and Uri's comment; please see my update.

versions/raml-10/raml-10.md
```diff
@@ -921,55 +932,32 @@ The ​**file**​ type can be used to constrain the content to send through for
 | minLength? | Specifies the parameter value's minimum number of bytes. Value MUST be equal or greater than 0. Defaults to 0
 | maxLength? | Specifies the parameter value's maximum number of bytes. Value MUST be equal or greater than 0. Defaults to 2147483647
 
-#### Enums
-
-All custom primitive types, except for the file type, can declare an "enum" property. It should contain a set of fixed values.
+### User-defined Facets
 
-Example usage of enums
+RAML provides a way to also declare user-defined restrictions (facets) that extend the predefined set of built-in restrictions for any scalar, array or object type. These can be defined using the `facets` property and its value is a map of user defined facets, specifically the value follows the syntax of a [property declaration](#property-declarations). Depending on whether a facet is optional or required, any subtype may or must define the facets that are declared in the type it extends. For example:
```
usarid: Let's replace this with:

Facets express various additional restrictions which types impose on their instances, such as the optional `minimum` and `maximum` facets for numbers, or the `enum` facet for scalars. RAML provides a way to declare additional user-defined facets for any data type. 

User-defined facets are defined using the OPTIONAL `facets` property on a type declaration. Its value is a map whose keys name the user-defined facets and whose corresponding values define the concrete values which the corresponding facet may take, following the same syntax as [property declaration](#property-declarations). That is, a facet declaration in the declaration of a type specifies the concrete values which subtypes may use for that facet to restrict instances of those subtypes. Note that this implies that user-defined facets on a type do not restrict instances of that type but rather only instances of subtypes of that type.

Facet names MUST NOT begin with open parentheses, to disambiguate them from annotations. User-defined facet names on a type MUST NOT match built-in facets on that type, nor facet names of any ancestor type in the type's inheritance chain.

If a facet of a type is declared as required, then any subtype of that type MUST define a value for that facet.

Here is an example which defines the capability to add restrictions on dates that indicate they may not fall on holidays:

versions/raml-10/raml-10.md
```diff
@@ -921,55 +932,32 @@ The ​**file**​ type can be used to constrain the content to send through for
...
-
-```yaml
-#%RAML 1.0
-title: My API With Types
-types:
-  Date:
-    type: string
-    facets:
-      format: string
+A RAML processor must follow specific rules when validating user-defined facets:
```
usarid: Let's replace this with:

Because user-defined facets are by definition not built into this RAML specification, and hence their semantic may not be understood by all RAML processors, a RAML processor may or may not choose to use user-defined facets on a type in validating instances of that type.

versions/raml-10/raml-10.md
```diff
@@ -921,55 +932,32 @@ The ​**file**​ type can be used to constrain the content to send through for
...
 
-  MyDate:
-    type: Date
-    format: 'dd.mm.yyyy'
-```
+* You can not start facet name with `(` (this is needed to avoid ambiguity with annotations).
+* You can not redeclare built-in facets; for example you can not declare facet with name `properties` in any type which inherits from object type.
+* You can not redeclare facets which has been defined in a super type. This also means that you can not inherit facets with the same name from two different types.
+* You can not validate an instance of a type against user-defined facets.
+* A RAML processors MAY be not required to validate an instance of a type against user-defined facets.
```
usarid: Remove this line.

versions/raml-10/raml-10.md
```diff
@@ -921,55 +932,32 @@ The ​**file**​ type can be used to constrain the content to send through for
...
+* A sub-type does not automatically inherit the user-defined facets from its parent type; hence a type that extends from `PossibleMeetingDate` must not define any of the user-defined facets that are part of `CustomDate`.
```
usarid: I don't understand exactly what this means. I would assume that instances of `PossibleMeetingDate` cannot fall on holidays, and that instances of any types that extend from `PossibleMeetingDate` also cannot fall on holidays, except in the case of union types which allow alternatives that do fall on holidays, e.g. 

```
types:
  ExtendedMeetingDate:
    type: PossibleMeetingDate | OffCalendarMeetingDate
```

But I cannot understand what you meant by
>A sub-type does not automatically inherit the user-defined facets from its parent type; hence a type that extends from `PossibleMeetingDate` must not define any of the user-defined facets that are part of `CustomDate`.

usarid:
See my comments, after which we're down to only one thing that needs clarification and resolution.

versions/raml-10/raml-10.md
sichvoge: All points or only the last?

usarid: I've replaced **all** bullet items one way or another, except for the one that starts with "A sub-type does not automatically inherit...". It's getting a bit hard to keep track of this PR's individual pieces, but you should be able to trace where every bullet point in your original PR went to, and see the final version emerge. Once we get clarity on that one remaining bullet point, we should be good to go.

sichvoge:
Updated; last thing left is the inheritance.

petrochenko-pavel-a:
> sub-type does not automatically inherit the user-defined facets from its parent type; hence a type that extends from PossibleMeetingDate must not define any of the user-defined facets that are part of CustomDate

Actually it looks like just rephrasing of the rule:

> You can not redeclare facets which has been defined in a super type. This also means that you can not inherit facets with the same name from two different types.

So I suggest just remove it.

sichvoge:
@petrochenko-pavel-a so you can inherit facets, but not facets with the same name or from union types, and you cannot redeclare facets that has been defined in the super type already?

sichvoge:
OK; since the text 

> Facet names MUST NOT begin with open parentheses, to disambiguate them from annotations. User-defined facet names on a type MUST NOT match built-in facets on that type, nor facet names of any ancestor type in the type's inheritance chain.

already mention all the different restriction, I removed all including the one with inheritance restrictions. 

@petrochenko-pavel-a, did I understood correctly that you can inherit facets, but not facets with the same name or from union types, and you cannot redeclare facets that has been defined in the super type already?

versions/raml-10/raml-10.md
```diff
@@ -440,6 +440,7 @@ declarations may have; certain type declarations may have other properties.
 | displayName? | An alternate, human-friendly name for the type
 | description? | A longer, human-friendly description of the type
 | (&lt;annotationName&gt;)? | Annotations to be applied to this type. Annotations are any property whose key begins with "(" and ends with ")" and whose name (the part between the beginning and ending parentheses) is a declared annotation name. See section on [Annotations](#annotations) for more information.
+| facets? | A map of user defined restrictions that are being inherited by a specific sub type. See section [User defined Facets](#user-defined-facets) for more information.
```

>A map of user defined restrictions that are being inherited by a specific sub type.

to

>A map of additional, user-defined restrictions that will be inherited and applied by any extending subtype.

versions/raml-10/raml-10.md
```diff
@@ -799,18 +800,28 @@ types:
 
 ### Scalar Types
 
-RAML provides a predefined set of Scalar Types. You can "extend" these types and add further restrictions. These restrictions are called "facets" and they exist both for scalar and object types. Each scalar type has a predefined set of facets.
+RAML defines a set of built-in scalar types and each of them has a predefined set of restrictions. All these types, except the file type, may have an additional `enum` property.
```
usarid: >may have an additional `enum` property

or

>may have an additional `enum` facet

?

versions/raml-10/raml-10.md
```diff
@@ -799,18 +800,28 @@ types:
...
 
 |Property Name | Description |
 |:--------|:------------|
-| facets? | When extending from a scalar type you can define new facets (which can then be set to concrete values by subtypes). The value is an object whose properties map facets names to their types.
-| enum? | Enumeration of possible values for this primitive type. Cannot be used with the file type. The value is an array containing string representations of possible values, or a single string if there is only one possible value.
+| enum? | Enumeration of possible values for this built-in scalar type. The value is an array containing string representations of possible values, or a single string if there is only one possible value.
```
usarid: Why string representations? Just an array containing possible values. (`1, 2, 3` are not string representations.)

usarid:
One comment, else LGTM

Issue #325, opened 2016-03-18
Add more clarification around the different default types and when they apply in which context


sichvoge:
The current RC1 version explains the different default types in section [Shortcuts and Syntactic Sugar](https://github.com/raml-org/raml-spec/blob/raml-10/versions/raml-10/raml-10.md#shortcuts-and-syntactic-sugar). In RC2 we should make sure that these are described a bit more clearly w/o using examples only.

There following rules need to be applied to determine the default type:

* If, and only if, a type declaration contains a `properties` node; the default type is `object`. For example (based on this rule):

```yaml
types:
   Person:
      type: object
      properties:
```

can also be written

```yaml
types:
   Person: 
      # default type is `object`, no need to explicitly define it
      properties:
```

* If, and only if, a type declaration does not contain either a `properties` node or no explicit `type`/`schema` node; the default type is always `string`. For example (based on this rule):

```yaml
types:
   Person:
      properties:
         name: # no type or schema necessary since the default type is `string`
```

* The default type `any` is applied to any `body` node regardless if it is a response or request declaration that does not contain `properties`, `types`, or `schema`. That rule is necessary since the value of a `body` node has the same structure like a type declaration except that the key is always a media type. For example:

```yaml
body:
  application/json: # key
    # default type is `any` 
```

The last rule is important for validation purposes and needs to be defined in the specification. Of course, each rule can be overridden by explicitly define a type. For example: 

```yaml
types:
   Person:
      properties:
         name: 
            type: number
```

PR #349, opened 2016-03-31
added rules for determining default types

sichvoge:
rules according to #325

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -971,7 +971,7 @@ types:
     properties:
 ```
 
-* If, and only if, a type declaration does not contain either a `properties` node or no explicit `type`/`schema` node; the default type is always `string`. For example (based on this rule):
+* If, and only if, a type declaration contains neither a `properties` node nor a `type` or `schema` node, then the default type is `string`. For example (based on this rule):
 
 ```yaml
 types:
@@ -995,7 +995,7 @@ body:
   # default type is `any`
 ```
 
-The last rule is important for validation purposes and needs to be defined in the specification. Of course, each rule can be overridden by explicitly define a type. For example:
+Of course, each rule can be overridden by explicitly define a type. For example:
 
 ```yaml
 types:
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -949,6 +949,55 @@ types:
     format: 'dd.mm.yyyy'
 ```
 
+### Determine Default Types
+
+A RAML processor must be able to determine the default type of a type declaration by using the following rules:
+
+* If, and only if, a type declaration contains a `properties` node; the default type is `object`. For example (based on this rule):
+
+```yaml
+types:
+  Person:
+    type: object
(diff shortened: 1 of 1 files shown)


petrochenko-pavel-a:
LGTM

versions/raml-10/raml-10.md
```diff
@@ -949,6 +949,55 @@ types:
...
+* If, and only if, a type declaration does not contain either a `properties` node or no explicit `type`/`schema` node; the default type is always `string`. For example (based on this rule):
+
+```yaml
+types:
+  Person:
+    properties:
+      name: # no type or schema necessary since the default type is `string`
+```
+
+* The default type `any` is applied to any `body` node regardless if it is a response or request declaration that does not contain `properties`, `types`, or `schema`. That rule is necessary since the value of a `body` node has the same structure like a type declaration except that the key is always a media type. For example:
```
usarid: First, `types` should be `type`.

Second, when you say "the key is always a media type", that's not explicitly true, only implicitly, because the media type may itself be defaulted, via the root-level `mediaType`. So I worry this statement is confusing. You can have:
```raml
body:
  type: string
```
And part of the confusion comes from this: "regardless if it is a response or request declaration that does not contain..." Can you separate out the pieces of that sentence so the rules are more explicit and obvious?

usarid:
Note changes requested in the comments.

versions/raml-10/raml-10.md
sichvoge: > First, `types` should be `type`.

So far, we used `types` more consistently than `type` inside the spec to outline one or multiple type definition(s). I'd stick with that.

sichvoge:
See last commit. It's not exactly like you said, but I simply removed the confusion. Let me know if that works or you really want to be much more explicit.

versions/raml-10/raml-10.md
```diff
@@ -949,6 +949,62 @@ types:
...
+can also be written
+
+```yaml
+types:
+  Person:
+    # default type is `object`, no need to explicitly define it
+    properties:
+```
+
+* If, and only if, a type declaration does not contain either a `properties` node or no explicit `type`/`schema` node; the default type is always `string`. For example (based on this rule):
```
usarid: Change the first sentence to state more clearly: 

If, and only if, a type declaration contains neither a `properties` node nor a `type` or `schema` node, then the default type is `string`.

versions/raml-10/raml-10.md
```diff
@@ -949,6 +949,62 @@ types:
...
+```
+
+Or in case that a default media type has been defined
+
+```yaml
+body:
+  # default type is `any`
+```
+
+The last rule is important for validation purposes and needs to be defined in the specification. Of course, each rule can be overridden by explicitly define a type. For example:
```
usarid: "The last rule is important for validation purposes and needs to be defined in the specification." <-- Why is this sentence here? Seems it should be gone.

usarid:
LGTM after incorporating 2 further comments.

PR #351, opened 2016-03-31
added new validation chapter with a section on how to validate type instances

### LINKED ISSUE #315: Add more clarifications and examples on how type instances get validated
RC1 spec does not specifies any rules for validation of type instances. RC2 should add more examples of instance validation as well as the rules associated with each facet. These rules will make clear what one should expect from a processor output.

sichvoge:
will close #315 but also prepares for other sections like how to validate type declarations etc.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -324,15 +324,9 @@ securitySchemes:
 
 ### Introduction
 
-RAML 1.0 introduces the notion of data **types**, which provide a concise and powerful way to describe the data in
-your API. The data can be in a URI parameter (base or resource URI), a query parameter, a request or response header,
-or of course a request or response body. Some types are built in, while custom types may be defined by extending
-(inheriting from) the built-in types. In any place where the API expects data, a built-in type may be used to describe
-the data, or a type may be extended inline to describe that data. Custom types may also be named and then used
-like any built-in type.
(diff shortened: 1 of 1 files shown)


usarid:
This is very strange, I don't know why we need this section at all. An instance of a type is valid when it validates against the type definition - it's tautological, but I think that's what you're saying. That's what 1) says; since 2) just deals with unions of types, which are themselves types, it's superfluous; and 3) should refer to validation rules for XML and JSON Schema (which version?) and not to libraries, but it belongs wherever we say that a type may be an XML or JSON Schema and may not be wrapped in, or extended by, a RAML data type.

sichvoge:
You are right by saying its tautological; I still think its quite helpful to understand the different concepts that plays a role in validating type instances. It basically calls out that instances will be validated against all restrictions (facets) of a type, or in case of union types that it needs to be valid against at least one type.  The third basically highlights that there are no special validation rules that the spec adds to the default validation rules for XML or JSON Schemas. Maybe calling out libraries is too much.

usarid:
No, I can't agree, in this case adding this section actually is confusing, because it makes you think: what am I missing that's not elsewhere? It's fine to say, in the introduction to types, that types are a way of restricting the data, and hence add rules for validating instances of data against the spec, so an instance is valid when it adheres to all rules for the type. That's all. As for union types, they are just types, so just like we define what's a valid string and what's a valid sequence, we should define in the union types section what's a valid instance of a union type.

sichvoge:
OK, I moved the "rules" into their respective chapters. Except for the external types since I believe it needs some rework anyways.

usarid:
LGTM

Issue #329, opened 2016-03-21
Add additional `xml` node for type declarations that allows configuring serialization of type instances to XML


sichvoge:
In RC1, the specification does not contain information about the serialization of type instances to XML. As the serialization to XML may be a complex process, RAML 1.0 RC2 introduces an additional `xml` node for type declarations that allows to configure how type instances should be serialised to XML.

The value of the `xml` node is a map that contains the following properties:

| Name | Type | Description |
|:---------|:------:|:-----------------|
| attribute? | `boolean` | If `attribute` is set to `true`, a type instance should be serialized as an XML attribute. It can only be `true` for scalar types.<br/><br/>**Default:** `false`
| wrapped? | `boolean` | If `wrapped` is set to `true`, a type instance should be wrapped in its own XML element. It can not  be `true` for scalar types and it can not  be `true` at the same moment when `attribute` is `true`. <br/><br/>**Default:** `false`
|  name? | `string` | Allows to override the name of the XML element or XML attribute in it's XML representation.<br/><br/>**Default:** the name of the type
| namespace? | `string` | Allows to configure the name of the XML namespace.
| prefix? | `string` |  Allows to configure the prefix which will be used during serialization to XML.

The following is a type declaration example that uses the `xml` node:

```yaml
types:
  Person:
    properties:
      name:
        type: string
        xml:
          attribute: true # serialize it as an XML attribute
          name: "fullname" # attribute should be called fullname
      addresses:
        type: Address[]
        xml:
          wrapped: true # serialize it into it's own <addresses>...</addresses> XML element
  Address:
    properties:
      street: string
      city: string
```

The example above may be serialized into the following XML:

```xml
<Person fullname="John Doe">
  <addresses>
     <Address>…</Address>
     ...
  </addresses>
</Person>
```

machaval:
Why aren't this xml specific things annotations in a Xml library? Are we going to extend the grammar for CSV, FixWidth etc? I think the way to scale it to have one libraries for each one so that we can show that the type system is extensible.

sichvoge:
Annotations are usually very practical for things that are not standardized yet, and also allow a lot flexibility and freedom of course. The goal here is to standardize a common use case by introducing a mechanism to serialize type instances into XML. This could be used across different tools. 

That does not necessarily mean you cannot use annotations and define a library if these, but this is by far the most common use case, as I said, and we want to make sure that there is a built-in behaviour or spec mandated behaviour.

Are you saying that it should not be built-in?

[...]

PR #354, opened 2016-04-04
added additional type declaration facet 'xml' and the respective chapter

sichvoge:
according to #329 any type declaration may define an additional `xml` node to allow the configuration of serialization of type instances into XML.


usarid:
LGTM

Issue #341, opened 2016-03-30
Are you allowed to pull in libraries in global definitions such as traits and resource types?


sichvoge:
Currently, the RC1 spec allows to define `uses` and therefore bringing in libraries inline. Are we really gonna allow that? If yes, why do would we allow that?

![image](https://cloud.githubusercontent.com/assets/498114/14138655/639f51a8-f666-11e5-9da5-9a07ab4f1750.png)
 
IMHO, that should not be allowed since I don't really see any use cases to do that on any global declaration.

usarid:
I thought we settled definitely that `uses` can only be applied at the root level of a document (RAML file or fragment), so it could not be used within the `traits` or `resourceTypes` declaration. Please clarify with @petrochenko-pavel-a

petrochenko-pavel-a:
@usarid @sichvoge as well as I remember `uses` is a top level property defined for any fragment now.

sichvoge:
Thanks for clarifying Pavel.

PR #355, opened 2016-04-04
remove 'uses' from any resource type or trait definition

sichvoge:
according to #341 `uses` is a root-level node and hence not allowed inside any other nodes.


usarid:
LGTM

PR #361, opened 2016-04-06
clarified the value of the 'types' node

### LINKED ISSUE #360: Clarify the value of the 'types' node
The value of the root-level `types` node is a map where each key represents the name of the type that can be referred to, and the value a type declaration.

sichvoge:
closes #360


usarid:
LGTM

Issue #364, opened 2016-04-07
Re-work "External Types" section


sichvoge:
The current [External Types](https://github.com/raml-org/raml-spec/blob/raml-10/versions/raml-10/raml-10.md#external-types) section can be confusing since the specification is actually talking about "Defining/Using XML and JSON Schema". We should reflect that in the title a bit more clear and remove occurrences of the term "external" that has been used to identify everything that is not the new RAML Data Type System.

We should also incorporate #314 into that as it directly relates to that chapter. So the following changes needs to be made:

* Change title to "Using XML and JSON Schema"
* Remove occurrences of "external" and replace it with something that refers better to other schema languages such as JSON and XML
* Wordsmithing the content of this chapter
* Incorporate #314

Issue #314, opened 2016-03-15
Remove any support for inline declaration of JSON Schema or XSD in any type declaration


sichvoge:
The problem here is that having the ability to define inline JSON Schema or XSDs in any type declaration may introduce a lot ambiguity and creates very complex use cases when serializing it into something more specific. Therefore, we suggest to remove the support for inline declaration of both in any type declaration. One will still be able to define XSDs, JSON Schema, or types; but is not able to mix them.

bademux:
why?
there are already format hel. Leave it as is or use yaml only, imho

freddrake:
This sounds like excessive hand-holding, better addressed by a best practices document that can evolve independently and include rationale.

There are plenty of times highly localized types are appropriate; whether those should be pulled out or not will often be determined by human judgement.  No need to require one approach or another in the spec.

sichvoge:
The biggest problem here is serialization of a specific type that mix up all three since not everything is interchangeable between these three formats. The questions that we would need to answer here are how these things are getting serialized into one specific which often is the case. How do you serialize it into XML after defining a type that mix all three?

It's something that we could not easily answer and therefore simplify the concepts for the type systems very first version. That does not mean we cannot evolve and bring these in when there is a better understanding how that will work, and how people use or want to use it.

bademux:
@sichvoge  I'm not sure I fully understand the problem, can you please provide an example?

sichvoge:
I only have a simple example that would illustrate what I mean by mixing types.

```yaml
  MixType:
    properties:
      JsonProperty:
        type: |
          {
            "title": "Example Schema",
            "type": "object",
            "properties": {
              "firstName": {
                "type": "string"
              },
              "lastName": {
                "type": "string"
              },
              "age": {
                "description": "Age in years",
                "type": "integer",
                "minimum": 0
              }
            },
            "required": ["firstName", "lastName"]
          }
      XmlProperty:
        type: |
          <xs:element name="item" maxOccurs="unbounded">
            <xs:complexType>
              <xs:sequence>
                <xs:element name="title" type="xs:string"/>
                <xs:element name="note" type="xs:string" minOccurs="0"/>
                <xs:element name="quantity" type="xs:positiveInteger"/>
                <xs:element name="price" type="xs:decimal"/>
              </xs:sequence>
            </xs:complexType>
          </xs:element>
      NormalType:
        properties:
          name: string
```

[...]

PR #365, opened 2016-04-07
full re-work of the 'External Types' chapter

sichvoge:
This PR contains a first draft of the re-worked "External Types" section. I made changes according to #364 and #314.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1189,7 +1189,7 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
 
 ### Using XML and JSON Schema
 
-RAML allows to not only define data using the type system, although it is recommended, but also provides mechanism to define them in XML and/or JSON schemas. RAML's type system allows seamless integration of JSON and XML schemas as type definitions.
+RAML allows the use of XML and JSON schema to describe the body of an API request or response, by integrating them into its data type system.
 
 The following examples show how to include an external JSON schema into a root-level type definition and a body declaration.
 
@@ -1241,9 +1241,9 @@ types:
       members: Person[] # invalid use of type expression '[]' and as a property type
  ```
 
-A RAML Processor MUST be able to interpret JSON Schema and XML Schema.
+A RAML Processor MUST be able to interpret and apply JSON Schema and XML Schema.
 
-XML schema MUST NOT be used where the media type does not allow XML-formatted data, and JSON schema MUST NOT be used where the media type does not allow JSON-formatted data. XML and JSON schemas are also FORBIDDEN in any declaration of query parameters, query string, and headers.
+XML schema MUST NOT be used where the media type does not allow XML-formatted data, and JSON schema MUST NOT be used where the media type does not allow JSON-formatted data. XML and JSON schemas are also forbidden in any declaration of query parameters, query string, URI parameters, and headers.
 
 Please note that the properties "schemas" and "types" are completely synonymous, so are "schema" and "type" for compatibility with RAML 0.8, but "schemas" and "schema" are deprecated. API definitions should use "types" and "type", as "schemas" and "schema" may be removed in a future RAML version.
 
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1208,7 +1208,7 @@ types:
             type: !include person.json
 ```
 
-A RAML processor MUST not allow types that define an XML or JSON schema to participate in type inheritance or specialization. In other words: You cannot define sub-types of these types that declare new properties, add restrictions, and set or declare facets. You can, however, create simple type wrappers that add annotations, examples or a description.
+A RAML processor MUST not allow types that define an XML or JSON schema to participate in type inheritance or specialization, or effectively in any [type expression](#type-expressions). In other words: You cannot define sub-types of these types that declare new properties, add restrictions, and set or declare facets. You can, however, create simple type wrappers that add annotations, examples or a description.
 
 The following is a fully valid example.
 
@@ -1228,6 +1228,18 @@ types:
     properties: # invalid
       single: boolean
 ```
+
+Another invalid case is the following example where the type `Person` is being used for a property type.
+
+```yaml
+types:
+  Person:
+    type: !include person.json
+    description: this is a schema describing person
+  Board:
+    properties:
+      members: Person[] # invalid use of type expression '[]' and as a property type
+ ```
 
 A RAML Processor MUST be able to interpret JSON Schema and XML Schema.
 
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1243,6 +1243,8 @@ types:
 
 A RAML Processor MUST be able to interpret JSON Schema and XML Schema.
 
+XML schema MUST NOT be used where the media type does not allow XML-formatted data, and JSON schema MUST NOT be used where the media type does not allow JSON-formatted data. XML and JSON schemas are also FORBIDDEN in any declaration of query parameters, query string, and headers.
+
 Please note that the properties "schemas" and "types" are completely synonymous, so are "schema" and "type" for compatibility with RAML 0.8, but "schemas" and "schema" are deprecated. API definitions should use "types" and "type", as "schemas" and "schema" may be removed in a future RAML version.
 
 ### Inheritance and Specialization
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1268,9 +1268,18 @@ types:
             type: !include person.json
 ```
 
-A RAML processor MUST not allow types that define an XML or JSON schema to participate in type inheritance or specialization. In other words: You cannot define sub-types of external types that declare new properties or set facets. You can, however, create simple type wrappers that add metadata, examples and a description.
+A RAML processor MUST not allow types that define an XML or JSON schema to participate in type inheritance or specialization. In other words: You cannot define sub-types of these types that declare new properties, add restrictions, and set or declare facets. You can, however, create simple type wrappers that add annotations, examples or a description.
 
-This example shows an invalid case where a type inherits the characteristics of a JSON schema and adds additional properties.
+The following is a fully valid example.
+
+```yaml
+types:
+  Person:
+    type: !include person.json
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,99 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
+/person:
+  get:
+    responses:
+      200:
+        body:
+          application/json:
+            type: !include person.json
+```
+
+A RAML processor MUST not allow types that define an XML or JSON schema to participate in type inheritance or specialization. In other words: You cannot define sub-types of external types that declare new properties or set facets. You can, however, create simple type wrappers that add metadata, examples and a description.
```
petrochenko-pavel-a: This needs more clarification: I would suggest the following wording:

You cannot define sub-types of external types that declare new properties add restrictions ,set or declare facets. You can, however, create simple type wrappers that add annotations, examples or  description

versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,99 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
+This example shows an invalid case where a type inherits the characteristics of a JSON schema and adds additional properties.
+
+```yaml
+types:
+  Person:
+    type: !include person.json
+    properties: # invalid
+      single: boolean
+```
+
```
petrochenko-pavel-a: I also would suggest to add valid example as well:

```yaml
types:
    Person:
      type: !include person.json
      description: this is a schema describing person
```

sichvoge:
Thanks @petrochenko-pavel-a; I applied your comments.

petrochenko-pavel-a:
LGTM

versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,108 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
   color: "brown"
 ```
 
+### Using XML and JSON Schema
+
+RAML allows to not only define data using the type system, although it is recommended, but also provides mechanism to define them in XML and/or JSON schemas.
```
usarid: Replace this line with:

RAML allows the use of XML and JSON schema to describe the body of an API request or response, in lieu of using RAML data types to describe that body.

versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,108 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
+
+A RAML Processor MUST be able to interpret JSON Schema and XML Schema.
```
usarid: Add that it MUST be able to validate instances of body data in XML and JSON formats against XML and JSON schema, respectively.

versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,108 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
+
+#### Schemas Node
```
usarid: Why do you put the `schemas` node here? It may indicate to users that `Schemas` should be used for XML and JSON schemas, whereas in fact, it is completely synonymous with `types` so there's no correlation between when you use `schemas` vs `types` and where you can declare XML and JSON schemas vs declaring RAML data types. Let's NOT seem to correlate them in the spec: describe the `schemas` and `schema` synonyms in the spec right after you introduce the `types` and `type` facets, respectively, say they're deprecated as we do here, perhaps give a simple example to show they are indeed synonyms (and don't correlate in the examples, i.e. please show `schemas` and `schema` used with RAML data types), and then don't mention `schemas` and `schema` in the spec again.

versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,108 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
+                  }
+                }
+              }
+  post:
+    body:
+      application/json:
+        schema: !include person.json
+```
+
+The **schema** node CANNOT be specified if a body's media type is application/x-www-form-urlencoded or multipart/form-data, or if a "type" node has been specified already (see [Using RAML Data Type System](#using-raml-data-type-system)).
```
usarid: This is odd. The fact that you cannot use both "schema" and "type" should have already been stated when we declared them as synonyms. And as for compatible and incompatible media types, there are infinitely many others that are also potentially compatible or incompatible, so why do we pick these? Perhaps we should simply state that XML schema MUST NOT be used where the media type does not allow XML-formatted data, and that JSON schema MUST NOT be used where the media type does not allow JSON-formatted data.

versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,108 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
+/person:
+  get:
+    responses:
+      200:
+        body:
+          application/json:
+            type: !include person.json
+```
+
+A RAML processor MUST not allow types that define an XML or JSON schema to participate in type inheritance or specialization. In other words: You cannot define sub-types of these types that declare new properties, add restrictions, and set or declare facets. You can, however, create simple type wrappers that add annotations, examples or a description.
```
usarid: Please also mention that XML and JSON schemas cannot participate in type expressions (as those effectively extend types), and verify that annotations/examples/description are the only things you may add in those wrappers.

versions/raml-10/raml-10.md
sichvoge: Sure, just used the same phrase that we had in 0.8.

versions/raml-10/raml-10.md
sichvoge: I am doing that to represent every node which I can easily link where its been mention (eg in the table as a value type like the example below). 

| Name | Value Type | Description |
|:---------|:----------------|:----------------|
| schemas | [Schema Node](#) | ....

If I would not do that, what would the link in the value type look like if its just mentioned somewhere as a side note?

versions/raml-10/raml-10.md
```diff
@@ -1189,88 +1189,37 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
-        schema: !include person.json
+            type: !include person.json
 ```
 
-The **schema** node CANNOT be specified if a body's media type is application/x-www-form-urlencoded or multipart/form-data, or if a "type" node has been specified already (see [Using RAML Data Type System](#using-raml-data-type-system)).
-
-Please note that "schema" is an alias for the equivalent "type" node, for compatibility with RAML 0.8 and is deprecated. API definitions should use the "type" node, as the "schema" alias may be removed in a future RAML version. The "type" node allows for XML and JSON schemas (see [Using RAML Data Type System](#using-raml-data-type-system)).
-
-#### Using RAML Data Type System
+A RAML processor MUST not allow types that define an XML or JSON schema to participate in type inheritance or specialization, or effectively in any [type expression](#type-expressions). In other words: You cannot define sub-types of these types that declare new properties, add restrictions, and set or declare facets. You can, however, create simple type wrappers that add annotations, examples or a description.
```
petrochenko-pavel-a: I think we still should write that you can not use xml/json schemas with multipart/form-data

versions/raml-10/raml-10.md
```diff
@@ -1189,88 +1189,37 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
-  post:
-    body:
-      application/json:
-        schema: !include person.json
+            type: !include person.json
 ```
 
-The **schema** node CANNOT be specified if a body's media type is application/x-www-form-urlencoded or multipart/form-data, or if a "type" node has been specified already (see [Using RAML Data Type System](#using-raml-data-type-system)).
-
-Please note that "schema" is an alias for the equivalent "type" node, for compatibility with RAML 0.8 and is deprecated. API definitions should use the "type" node, as the "schema" alias may be removed in a future RAML version. The "type" node allows for XML and JSON schemas (see [Using RAML Data Type System](#using-raml-data-type-system)).
```
petrochenko-pavel-a: QQ: why you removed it?

sichvoge:
@usarid to your comment

> XML schema MUST NOT be used where the media type does not allow XML-formatted data, and that JSON schema MUST NOT be used where the media type does not allow JSON-formatted data.

Do we also need to clarify that you should not be allowed to use XML or JSON schemas for headers, query parameters, and query string? Is that even a correct assumption? Something like:

```yaml
uriParameters: 
    name: |
      {
        ....
      }
```

sichvoge:
OK, just got a confirmation seeing someone asking the same question for 0.8 and no its not supported. I'll update the PR and you can review and comment.

sichvoge:
See new commit.

versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,66 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
   color: "brown"
 ```
 
+### Using XML and JSON Schema
+
+RAML allows to not only define data using the type system, although it is recommended, but also provides mechanism to define them in XML and/or JSON schemas. RAML's type system allows seamless integration of JSON and XML schemas as type definitions.
```
usarid: Change the entire line (both sentences) to:

>RAML allows the use of XML and JSON schema to describe the body of an API request or response, by integrating them into its data type system.

versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,66 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
+types:
+  Person:
+    type: !include person.json
+    description: this is a schema describing person
+  Board:
+    properties:
+      members: Person[] # invalid use of type expression '[]' and as a property type
+ ```
+
+A RAML Processor MUST be able to interpret JSON Schema and XML Schema.
```
usarid: change "to interpret" to "to interpret and apply"

versions/raml-10/raml-10.md
```diff
@@ -1187,6 +1187,66 @@ CatOrDog: # follows restrictions applied to the type 'Cat'
...
+
+XML schema MUST NOT be used where the media type does not allow XML-formatted data, and JSON schema MUST NOT be used where the media type does not allow JSON-formatted data. XML and JSON schemas are also FORBIDDEN in any declaration of query parameters, query string, and headers.
```
usarid: also URI parameters

usarid:
See my small comments, otherwise LGTM.

Issue #318, opened 2016-03-16
finalize all use of data types in named parameters


sichvoge:
In RC1, a named parameter such as header or query parameter is defined as a "properties declaration". That might impose restrictions, for example, between query parameters. We need to finalize what the value of named parameters are in relation with types.

usarid:
Replace the entire contents of the section titled "The Query String as a Whole" with the following, after making it consistent, if necessary, with our new terminology for describing YAML nodes, and after making the sentence containing "must be derived from" consistent with the way we discuss type expressions and extending types.

Note that this example is the same as in #319.

---

The queryString property is used to specify the query string as a whole, rather than as name-value pairs. Its value is either the name of a type or an inline type declaration. The type must be derived from a scalar type or from a map type.

If the type is derived from a scalar type, the query string as a whole MUST be described by the type. 

If the type is derived from a map type, processors MUST regard the query string as a URL-encoded serialization of an instance of this map type; that is, the query string must be of the form "parameter1=value1&parameter2=value2&..." where "parameter1", "parameter2", etc. correspond to the keys in the map type, and the values to the corresponding value specifications in the map type. If a value in the map type is an array type, processors MUST interpret this as allowing multiple instances of that query parameter in the query string. In such a case, the underlying type of the array -- namely, the type of the elements of the array -- MUST be applied as the type of the value of instances of this query parameter.

In the following example, union types and extending from multiple types are used to constrain the query parameters to specific alternatives:
```raml
#%RAML 1.0
title: Illustrate query parameter variations
types:
  lat-long: # lat & long required; mutually exclusive with location
    properties:
      lat: number
      long: number
  loc: # location required; mutually exclusive with lat & long
    properties:
      location:
  paging: # each is optional, not exclusive with anything
    properties:
      start?: number
      page-size?: number
/locations:
  get:
    queryString:
      type: [paging,  lat-long | loc ]
      examples:
        first:
          content:
            start: 2
            lat: 12
            long: 13
        second:
          content:
            start: 2
            page-size: 20
            location: 1,2
        third:  # not valid
          content:
            lat: 12
            location: 2
          strict: false # because it's not valid
```

usarid:
I don't know that the other section on named parameters (URI parameters, base URI parameters, and headers) need any changes.

sichvoge:
Don't think so since most clearly state that the value of each is equal to a `properties` declaration. If we find something else, we can easily create a separated issue and clean that.

PR #366, opened 2016-04-07
changed content of the 'queryString' description 

sichvoge:
according to #318

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1904,7 +1904,7 @@ An API's methods may support or require a query string in the URL on which they 
 
 #### The Query String as a Whole
 
-The queryString property is used to specify the query string as a whole, rather than as name-value pairs. Its value is either the name of a type, including union types; or an inline type declaration. The type must be derived from a scalar type or from an object type.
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -1904,20 +1904,49 @@ An API's methods may support or require a query string in the URL on which they
 
 #### The Query String as a Whole
 
-The queryString property is used to specify the query string as a whole, rather than as name-value pairs. Its value is either the name of a type or an inline type declaration.
+The queryString property is used to specify the query string as a whole, rather than as name-value pairs. Its value is either the name of a type, including union types; or an inline type declaration. The type must be derived from a scalar type or from an object type.
```
usarid: "the name of a type, including union types" should probably be "the name of a data type, or an inline data type declaration (e.g. a data type expression)".

usarid: We should also probably clarify something to the effect of: "The type must be derived from a scalar type or from an object type, after resolving all data type expressions." @petrochenko-pavel-a can you please make sure this is the right way to state it?

usarid:
See my changes, and if they're OK then LGTM.

versions/raml-10/raml-10.md
usarid: OK let's change the entire thing to:

The queryString property is used to specify the query string as a whole, rather than as name-value pairs. Its value is either the name of a data type or an inline data type declaration (including a data type expression). In either case the types at the root of the type hierarchy of the data type MUST all be either a scalar type or the object type, after fully expanding any union type expressions at every level of the type hierarchy.

Issue #367, opened 2016-04-07
Clarify that how multiple fully-resolved paths that are identical are handled


sichvoge:
According to https://github.com/raml-org/raml-spec/issues/184#issuecomment-200986680; the spec should indicate that multiple fully-resolved paths that are identical (up to character encoding) are forbidden, and otherwise they are permitted.

PR #368, opened 2016-04-07
made more explicit that RAML does not allow multiple paths to be resolved into the same

sichvoge:
According to #367

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1671,6 +1671,26 @@ https://api.github.com/users/{userId}/keys
 https://api.github.com/users/{userId}/keys/{keyId}
 ```
 
+A RAML processor MUST not allow a fully-combined path, which you get from combining the resource's path with all of its parents, to be identical to another fully-combined path, WITHOUT consideration to the possible values of any URI parameter.
+
+This example would be FORBIDDEN.
+
+```yaml
+/users:
+  /foo:
+/users/foo:
+```
+
+(both paths combine to the same `/users/foo`), and on the other hand this would ALWAYS be ALLOWED.
+
+```yaml
+/users/{userId}:
+/users/{username}:
+/users/me:
+```
+
+### Resource Property
+
 The value of a resource property is an object whose properties are described in the following table.
 
 |Property | Description |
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -1671,6 +1671,26 @@ https://api.github.com/users/{userId}/keys
 https://api.github.com/users/{userId}/keys/{keyId}
 ```
 
+A RAML processor MUST not allow a fully-combined path, which you get from combining the resource's path with all of its parents, to be identical to another fully-combined path, WITHOUT consideration to the possible values of any URI parameter.
```
usarid: MUST not --> MUST NOT

usarid: We should find a better way to say it. 

First, define a path expansion algorithm. I think we already have that, in the definition of deriving the absolute URI for each resource -- but we should tweak that section so it does not imply you have to have a `baseUri` to proceed, and gives you a path expansion algorithm you can apply with or without a `baseUri`.

Once you're done that, refer to that algorithm here, and say that two fully-resolved paths cannot be identical, when they're compared with each other without resolving (substituting the values of) any URI parameters.

usarid:
See comments.

sichvoge:
See new commit.

usarid:
Needs more changes per our Slack

versions/raml-10/raml-10.md
```diff
@@ -1622,7 +1622,7 @@ baseUri: https://api.github.com
 
 The key of a resource property, i.e. its relative URI, MAY consist of multiple URI path fragments separated by slashes; e.g. /bom/items may indicate the collection of items in a bill of materials as a single resource. However, if the individual URI path fragments are themselves resources, the API definition SHOULD use nested resources to describe this structure; e.g. if /bom is itself a resource then /items should be a nested resource of /bom, vs using /bom/items as a non-nested resource.
 
-Absolute URIs are not explicitly specified. They are computed by starting with the baseUri and appending the relative URI of the top-level resource, and then successively appending the relative URI values for each nested resource until the target resource is reached.
+Absolute URIs are not explicitly specified. They are computed by appending the relative URI of the top-level resource, and then successively appending the relative URI values for each nested resource until the target resource is reached. If and only if the baseUri is defined, the baseUri is appended to the first segment, which represents the top-level resource, of relative URIs.
```
>If and only if the baseUri is defined, the baseUri is appended to the first segment, which represents the top-level resource, of relative URIs.
to
>In this formation of the absolute URI, if a baseUri is defined, it is prepended before the relative URI of the top-level resource; any trailing slashes in the baseUri are removed before prepending.

versions/raml-10/raml-10.md
```diff
@@ -1671,6 +1671,26 @@ https://api.github.com/users/{userId}/keys
 https://api.github.com/users/{userId}/keys/{keyId}
 ```
 
+A RAML processor MUST NOT allow one of the computed absolute URIs to be identical to another one, without consideration to the possible values of any URI parameter.
```

>A RAML processor MUST NOT allow one of the computed absolute URIs to be identical to another one, without consideration to the possible values of any URI parameter.

to

>A RAML processor MUST NOT allow one of the computed absolute URIs to be identical to another one; comparison of absolute URIs is done without consideration to the possible values of any URI parameter, i.e. any URI parameters are not expanded or evaluated but rather left as is.

usarid:
See my comments, otherwise LGTM.

Issue #370, opened 2016-04-08
Clarify array types to be a valid definition for `type`


sichvoge:
Currently the spec does not clearly states that `array` is a valid value for a `type` node. Its basically just stating that one can define an array type by using the array qualifier at the end of a type expression.

So following are both valid examples:

```yaml
types:
  Email:
    type: object
    properties:
      name:
        type: string
  emails:
    type: Email[]
    minItems: 1
    uniqueItems: true
```

```yaml
types:
  Email:
    type: object
    properties:
      name:
        type: string
  emails:
    type: array
    items: Email
    minItems: 1
    uniqueItems: true 
```

Using `Email[]` is equivalent to use `type: array` and specifically define the type items in the array inherit from using `items` facet.

PR #371, opened 2016-04-08
added more clarity about the array type

sichvoge:
according to #370; how to use the array type was not clear and created some confusion about using it as a direct type value w/o the array qualifier `[]`. That PR changes that.


fededonna:
LGTM :+1:

Issue #372, opened 2016-04-08
Modify examples


fededonna:
``` yaml
#%RAML 1.0 Overlay
usage: Hints for monitoring the library books API
extends: librarybooks.raml
annotationTypes:
  monitor:
    parameters:
      frequency:
        properties:
          interval: integer
          unitOfMeasure:
            enum: [ seconds, minutes, hours ]
      script:
/books:
  get:
    (monitor):
      frequency:
        interval: 5
        unitOfMeasure: minutes
      script: randomBooksFetch
```

``` yaml
#%RAML 1.0
title: Example API
version: v1
traits:
  parameters:
    queryParameters:
      platform:
        enum:
          - win
          - mac
/installer:
  get:
    queryParameters:
      platform: #the actual enum is [ mac, unix, win ]
        enum:
          - mac
          - unix
```

both examples have a "parameters" field which is not valid.

``` yaml
#%RAML 1.0
title: Illustrating annotations
mediaType: application/json
annotationTypes:
  experimental:
  feedbackRequested:
  testHarness:
    type: string # This line may be omitted as it's the default type
  badge:         # This annotation type, too, allows string values
  clearanceLevel:
    properties:
      level:
        enum: [ low, medium, high ]
        required: true
      signature:
        pattern: "\\d{3}-\\w{12}"
        required: true
/groups:
  (experimental):
  (feedbackRequested):
/users:
  (testHarness): usersTest
  (badge): tested.gif
  (clearanceLevel):
    level: high
    signature: 230-ghtwvfrs1itr
  get:
    (experimental):
    (feedbackRequested):
    responses:
      200:
```

when the annotation is used it should have a value because the lack of it is like putting a null node.

sichvoge:
Can you send a PR for this @fededonna (please use the `raml-10-rc2` branch.

usarid:
Agreed that `parameters` is out of place in something that should act like a property declaration, i.e. have things like `type` or `properties` in it.

But for your second claim, annotations do not need to have a value.

sichvoge:
This is related to #394.

sichvoge:
Actually closing that. The null-valued annotations are handled with #394

PR #380, opened 2016-04-12
Modify examples according to #372

fededonna:
according to #372

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -2431,7 +2431,7 @@ All the collections or sequences which fall under effect of applying traits and 
 title: Example API
 version: v1
 traits:
-  parameters:
+  withQueryParameters:
     queryParameters:
       platform:
         enum:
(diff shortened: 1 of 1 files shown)


sichvoge:
I pull in @usarid and @petrochenko-pavel-a to confirm what the default value for an empty annotation actually is. I can see use cases where you don't want to have a value, but rather just the annotation.

The initial case was that the example still used `parameters`, but @fededonna also outline that values for an annotations should not be empty since it creates it with `null`.

usarid:
Annotations can certainly be specified with no values if desired. What's wrong with `null`?

usarid:
We'll probably do something along the lines of https://github.com/raml-org/raml-spec/issues/394 now, to add a `null` type and therefore allow us to have valueless annotations after all.

@sichvoge in this case we can set the annotation type of `experimental` and `feedbackRequested` to be `null` in which case that's the only value they can take. Or you can set one to be `null` and the other `null | string` to illustrate some options. Your call.

sichvoge:
OK, I'll wait for discussion on #394 to be finished before changing it.

sichvoge:
Fixed issues in PR #409. This PR will no longer be merged, but leave it open bc there is still an open discussion.

sichvoge:
Closing that PR and create a new one when #409 is done!

PR #385, opened 2016-04-15
Issue323

krishahn:
Issue 323 and copyedits


sichvoge:
close that one in favour for #386

PR #393, opened 2016-04-17
copyedits, consistent highlighting of properties

krishahn:
Properties are enclosed in quotation marks, back ticks, and bold. In this section, the inconsistency was easy to spot, so I picked the most-used style of the surrounding text, quotation marks. This might not be the best choice. In technical writing, quotation marks are usually not used in this way and the prevalence of quotation marks in JSON might justify using back ticks/code formatting (least busy) or bold. A pass through the whole spec at some point just to make the highlighting consistent might be a good idea.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -414,8 +414,7 @@ RAML Types in a nutshell:
 
 ### Type Declarations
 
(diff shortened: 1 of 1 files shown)


sichvoge:
incorporated these changes into a new PR. closing that.

PR #395, opened 2016-04-18
Object Types copyedits

krishahn:
Editorial suggestions for the Object Types section. Some details in the properties table about the discriminator, such as limitations, do not appear in the Using Discriminator section.


sichvoge:
LGTM

Issue #407, opened 2016-04-18
 Re-work 'Multiple Inheritance' chapter.


sichvoge:
In RC1, multiple inheritance was restricted to object types only. That is not the case in RC2 anymore as long as the resulting type is a fully valid type.

The following example is valid:

```yaml
types:
   Number1:
     type: number
     minimum: 4
   Number2:
     type: number
     maximum: 10
   Number3: [ Number1, Number2]
```

Where as using the same example and only change the maximum value of type `Number2` from 10 to 2 would result into an invalid type `Number3`.

```yaml
types:
   Number1:
     type: number
     minimum: 4
   Number2:
     type: number
     maximum: 2
   Number3: [ Number1, Number2]
```

PR #410, opened 2016-04-18
wordsmithing the multiple inheritance chapter

sichvoge:
according to #407

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1232,7 +1232,7 @@ types:
 
 In the example above, the type `Teacher` inherits all restrictions from `Person` and `Employee`.
 
-Multiple inheritance is only allowed if the sub-type is still a valid type declaration after inheriting all restrictions from its parent types. For example, the type `Number3` is fully valid:
+Multiple inheritance is allowed only if the sub-type is still a valid type declaration after inheriting all restrictions from its parent types. For example, the type `Number3` is valid:
 
 ```yaml
 types:
@@ -1245,7 +1245,7 @@ types:
   Number3: [ Number1, Number2]
 ```
 
-Whereas using the same example and only change the maximum value of type `Number2` from 10 to 2 would result into an invalid type `Number3`.
+Whereas using the same example and only changing the maximum value of type `Number2` from 10 to 2 would result in an invalid type `Number3`.
 
 ```yaml
 types:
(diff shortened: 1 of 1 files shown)


krishahn:
Editorial suggestions committed.

usarid:
Now that we've removed the restriction on multiple inheritance being only for object types, we must specify whether you can have multiple inheritance from types that aren't extensions of the same built-in types, e.g. `type: [ number, string ]` or `type: [ number, integer ]` or `type: [ object, string ]` or `type: [ (string | boolean), string ]` or `type: [ foo, bar ]` where `foo` and `bar` are user-defined types of various sorts. @petrochenko-pavel-a can help clarify?

petrochenko-pavel-a:
>whether you can have multiple inheritance from types that aren't extensions of the same built-in types
There is no need to specially restrict it. But this types will be invalid, because they will have contra-dictional restrictions on they instanced for example `[number, boolean]` will have a restriction that instance should be javascript number and boolean at same moment

sichvoge:
Is there any general rule we need to apply or mention? @petrochenko-pavel-a w

petrochenko-pavel-a:
The generic rules is that type is invalid if it has any conflicting restrictions.

sichvoge:
> conflicting restrictions
maybe a very dumb question, but what does that mean exactly?

sichvoge:
OK so referring to that I'd suggest to call out the restrictions that you are not allowed to have different primitives participating in multiple inheritance.

sichvoge:
See last commit.

versions/raml-10/raml-10.md
```diff
@@ -1227,26 +1227,53 @@ When referencing an inner element of a schema, a RAML processor MUST validate an
...
+    properties:
+      employeeNr: integer
+  Teacher:
+    type: [ Person, Employee ]
 ```
 
-Note: Multiple inheritance is only allowed if all Type Expressions are simple object Types.
+In the example above, the type `Teacher` inherits all restrictions from `Person` and `Employee`.
+
+Multiple inheritance is allowed only if the sub-type is still a valid type declaration after inheriting all restrictions from its parent types. Also, it is not allowed to inherit from different kind of primitive types, for example `[ number, string ]`.
```
sichvoge: Uri's comment:

> I think you also need to explicitly state how to treat this statement when you have union types somewhere in the inheritance chain. I would guess the answer is to expand all the unions and and see that at least one of the allowed combinations is a valid type declaration and in particular does not combine different primitive types, but @petrochenko-pavel-a should say definitively.

usarid: I'm fine as long as it appears somewhere in the spec; in that case, you might want to just link to it from here.

PR #411, opened 2016-04-19
Several improvements for section "References to inner elements of external files"

krishahn:
copyedits of "References to Inner Elements of External Files"

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1212,6 +1212,19 @@ XML schema MUST NOT be used where the media type does not allow XML-formatted da
 
 Please note that the properties "schemas" and "types" are completely synonymous, so are "schema" and "type" for compatibility with RAML 0.8, but "schemas" and "schema" are deprecated. API definitions should use "types" and "type", as "schemas" and "schema" may be removed in a future RAML version.
 
+#### References to Inner Elements
+
+Sometimes it is necessary to refer to an element defined in a schema. RAML supports that by using URL fragments like in the example below.
+
+```yaml
+type: !include elements.xsd#Foo
+```
+
+Referencing an inner element of a schema is equivalent to referencing a top level element. The structural restrictions on referencing an inner element are the same as those on referencing other elements. This version of RAML specification supports referencing any inner elements in JSON schemas that are valid schemas, and any globally defined elements and complex types in XML schemas. There are only a few restrictions:
+
+* Validation of any XML or JSON instance against inner elements follows the same restrictions than the validation against a regular XML or JSON schema.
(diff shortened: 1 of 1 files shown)


sichvoge:
@krishahn I will add a bit more caviar and improvements to the section according to some clarifications we need to make. I'll also need to move that chapter into "Using XML and JSON Schema" section.

usarid:
>I will add a bit more *caviar* and improvements

Either autocorrect is having fun or we're really going upscale here

versions/raml-10/raml-10.md
```diff
@@ -1212,6 +1212,19 @@ XML schema MUST NOT be used where the media type does not allow XML-formatted da
...
 
+#### References to Inner Elements
+
+Sometimes it is necessary to refer to an element defined in a schema. RAML supports that by using URL fragments as shown in the example below.
+
+```yaml
+type: !include elements.xsd#Foo
+```
+
+Referencing an inner element of a schema is equivalent to referencing a top level element. The structural restrictions on referencing an inner element are the same as those on referencing other elements. This version of RAML specification supports referencing any inner elements in JSON schemas that are valid schemas, any globally defined elements, and complex types in XML schemas. There are only a few restrictions:
```
usarid: I don't know what it means to say "Referencing an inner element of a schema is equivalent to referencing a top level element." -- It could mislead people into thinking that you're referencing the entire schema for validation, which would be incorrect.

usarid:
One comment, else LGTM

sichvoge:
See last commit.

versions/raml-10/raml-10.md
```diff
@@ -1212,6 +1212,19 @@ XML schema MUST NOT be used where the media type does not allow XML-formatted da
...
 
+#### References to Inner Elements
+
+Sometimes it is necessary to refer to an element defined in a schema. RAML supports that by using URL fragments as shown in the example below.
+
+```yaml
+type: !include elements.xsd#Foo
+```
+
+When referencing an inner element of a schema, a RAML processor MUST validate an instance against that particular element. The structural restrictions on referencing an inner element are the same as those on referencing other elements. This version of RAML specification supports referencing any inner elements in JSON schemas that are valid schemas, any globally defined elements, and complex types in XML schemas. There are only a few restrictions:
```
usarid: >The structural restrictions on referencing an inner element are the same as those on referencing other elements.
I'm not sure what this means. Is it the same as the bullet point below, namely
>Validation of any XML or JSON instance against inner elements follows the same restrictions as the validation against a regular XML or JSON schema
? In that case, why repeat?

usarid:
One last comment, otherwise LGTM, so feel free to accept it yourself after resolving.

Issue #412, opened 2016-04-19
Make `types` and `schemas` mutual exclusive


sichvoge:
The spec does not clearly mention that `types` and `schemas` are also mutual exclusive. this is directly derived from the fact that `type` and `schema` are mutual exclusive, so `types` and `schemas` should also be. All these are synonymous, whereas `schemas` and `schema` should still be handled as "deprecated". Its is RECOMMENDED to use `types` and `type` in any case.

PR #415, opened 2016-04-19
make types and schemas mutual exclusive

sichvoge:
according to #412

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -155,6 +155,8 @@ The following table enumerates the possible properties at the root of a RAML doc
 | uses? | Imported external [libraries](#libraries) for use within the API.
 | /&lt;relativeUri&gt;? | The resources of the API, identified as relative URIs that begin with a slash (/). A [resource property](#resources-and-nested-resources) is one that begins with the slash and is either at the root of the API definition or a child of a resource property. For example, /users and /{groupId}.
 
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -155,6 +155,8 @@ The following table enumerates the possible properties at the root of a RAML doc
 | uses? | Imported external [libraries](#libraries) for use within the API.
 | /&lt;relativeUri&gt;? | The resources of the API, identified as relative URIs that begin with a slash (/). A [resource property](#resources-and-nested-resources) is one that begins with the slash and is either at the root of the API definition or a child of a resource property. For example, /users and /{groupId}.
 
+The "schemas" and "types" nodes are mutually exclusive and synonymous: processors MUST NOT allow both to be specified, explicitly or implicitly, at the root-level of an API definition. We recommended using the "types" node instead of "schemas" because the schemas alias is deprecated and might be removed in a future RAML version. Also, the "types" node supports XML and JSON schemas.
```
usarid: Why do you say "explicitly or implicitly" -- how can it be implicit?

And I would omit "Also, the "types" node supports XML and JSON schemas." -- it could mislead people into thinking there is a difference between them.

usarid:
See comments, otherwise LGTM

versions/raml-10/raml-10.md
sichvoge: Its explicitly indeed. Thanks!

PR #417, opened 2016-04-19
copyedit Object Type Specialization

krishahn:
copyedit of Object Type Specialization section

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -636,7 +636,7 @@ types:
         type: string
 ```
 
-A sub-type can override properties of its parent type with the following restrictions: 1) a required property in the parent type cannot become optional in the sub-type, and 2) the type declaration of a defined property in the parent type cannot widen in the sub-type. Changes that narrow the type (specialize the parent type) are allowed.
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -636,7 +636,7 @@ types:
         type: string
 ```
 
-Any sub-types can override properties defined in its parent types with the following two restrictions: 1) if a property is required in the parent type, it cannot be made optional in the sub-type, and 2) the type of an already defined property can only be changed to a narrower type (a type that specializes the parent type).
+A sub-type can override properties of its parent type with the following restrictions: 1) a required property in the parent type cannot become optional in the sub-type, and 2) the type declaration of a defined property in the parent type cannot widen in the sub-type. Changes that narrow the type (specialize the parent type) are allowed.
```
sichvoge: Is it "in the parent type cannot 'be' widen"?

krishahn:
Clarified, wordsmithing, thanks.

krishahn:
Stating the narrow instead of the wide case does work better and is succinct. Thanks.

krishahn:
one last massaging to recapture the feel of the original text

sichvoge:
LGTM, thanks @krishahn

PR #425, opened 2016-04-21
copyedits of User-defined facets plus link text correction

krishahn:
Copyedit User-defined Facets section and correct hyphenation of the link text in a previous section.


sichvoge:
LGTM

Issue #394, opened 2016-04-17
Support for null values


petrochenko-pavel-a:
It seems that RAML data types do not provide a way to allow for null-valued data in payloads, null-valued annotations, etc. 

An example of that is 
```raml
type:
  properties:
    name:
    comment:
example:
  name: Fred
  comment:
```
Here the type is object-valued with 2 required properties, `name` and `comment`, both defaulting to be of type `string`, and the example indeed gives both but `comment` has value of `null`. Since `null` is not of type `string` in YAML, it's not a valid value for this example, and there's no way to make it valid. One could use `""` (the empty string) as the value, but there should be a difference between a value that's `null` and a value that is a string of length 0. In many popular payload formats, such as JSON, there is such a difference, and RAML should be able to represent that easily. (Note also that the `required` facet only requires the key to be present, not the value to be non-null.)
 
**Proposed solution**: Lets add a `null` type to the list of built-in RAML scalar types. One would then be able to represent nullable properties as a union type of their normal type and `null` type. For the example above, one would use
```raml
type:
  properties:
    name:
    comment: string | null
example:
  name: Fred
  comment:
```

fededonna:
I think this is not right. This is a trick just for annotations. We should do a change in the spec and say:

Values are not required in scalar annotations, so they allow an empty value in the key value node. Just for annotations.

sichvoge:
I think that should even be more restricted to `string` only right? Since you can't really have an empty value for integer or number? For `string` that would be easier.

fededonna:
Ok, only for strings sounds good.

usarid:
@fededonna: I don't see why you're saying we should do something special for annotations vs introducing the `null` type and addressing the more general issue of nullable values, eg in JSON payloads.

sichvoge:
Do you have an example @usarid? I can't think about one that would need to be set `null`.

usarid:
JSON, for example, allows properties (or array elements) to be null; that's one of its scalar types. How do you describe a payload that can have a null in it without introducing `null` into RAML data types? For an example, see the description in this issue. Express the example in JSON if you wish.

sichvoge:
OK, I see; missed that even JSON defines a `null` type (see http://spacetelescope.github.io/understanding-json-schema/reference/type.html)

3miliano:
[...]

PR #428, opened 2016-04-21
added null type to the spec 

sichvoge:
according #394

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -916,6 +916,45 @@ The ​**file**​ type can constrain the content to send through forms. When th
 | minLength? | Specifies the minimum number of bytes for a parameter value. The value MUST be equal to or greater than 0.<br /><br />**Default:** `0`
 | maxLength? | Specifies the maximum number of bytes for a parameter value. The value MUST be equal to or greater than 0.<br /><br />**Default:** `2147483647`
 
+### Null Type
+​
+RAML handles null data values in payloads, annotations, and other constructs. Nullable properties must be represented as a single `null` type, a union of the non-null and `null` types, or suffixing the type with a trailing question mark `?`.
+In the following example, the type of an object and has two required properties, `name` and `comment`, both defaulting to type `string`. In `example`, `name` is assigned a string value, but comment is null and this is _not_ allowed because RAML expects a string.
+```yaml
+type:
+  properties:
+    name:
+    comment:
+example:
+  name: Fred
+  comment: # Providing no value here is not allowed.
+```
+The following example shows the assignment of the `null` type to `comment`:
+​
+```yaml
+type:
+  properties:
+    name:
+    comment: null
+example:
+  name: Fred
+  comment: # Providing a value here is not allowed.
+```
+The following example shows how to represent nullable properties using a union:
+​
+```yaml
+type:
+  properties:
+    name:
+    comment: string | null # equivalent to ->
+                           # comment: string?
+example:
+  name: Fred
+  comment: # Providing a value or not providing a value here is allowed.
+```
+
+Declaring the type of a property to be `null` causes a RAML processor to forbid an instance to have a value.
+
 ### User-defined Facets
 
 Facets express various additional restrictions beyond those which types impose on their instances, such as the optional `minimum` and `maximum` facets for numbers, or the `enum` facet for scalars. In addition to the built-in facets, RAML provides a way to declare user-defined facets for any data type.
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -916,6 +916,52 @@ The ​**file**​ type can constrain the content to send through forms. When th
 | minLength? | Specifies the minimum number of bytes for a parameter value. The value MUST be equal to or greater than 0.<br /><br />**Default:** `0`
 | maxLength? | Specifies the maximum number of bytes for a parameter value. The value MUST be equal to or greater than 0.<br /><br />**Default:** `2147483647`
 
+### Null Type
+​
+RAML handles null data values in payloads, annotations, and other constructs. Nullable properties must be represented as a single `null` type, a union of the non-null and `null` types, or by suffixing the type with a trailing question mark `?` which is the equivalent to a union between `null` and a type, for example: `null | number`.
```
usarid: This is actually a bit confusing, as you're trying to capture all the ways to end up with allowing a `null` and that's an arbitrarily long list. Instead, just focus on the `null` type being a scalar type and matching only null data, which in JSON is JSON's `null` (and maybe state the XML one too, not sure). Then standard serialization and deserialization rules will apply to it as they do to any other type, and there's no reason to say them. But it is helpful to simply illustrate how to use this, with a couple of examples that have e.g. union type expressions in them.

versions/raml-10/raml-10.md
```diff
@@ -916,6 +916,52 @@ The ​**file**​ type can constrain the content to send through forms. When th
...
+  properties:
+    name:
+    comment: null | string # equivalent to ->
+                           # comment: string?
+example:
+  name: Fred
+  comment: # Providing a value or not providing a value here is allowed.
+```
+
+Declaring the type of a property to be `null` causes a RAML processor to forbid an instance to have a value. In a RAML context that requires *values* of type `null` (vs just type declarations), the usual YAML `null` is used, e.g. when the type is `number | null` you may use `enum: [ 1, 2, ~ ]` or more explicitly/verbosely `enum: [ 1, 2, !!null "" ]`; in non-inline notation you can just omit the value completely, of course.
```
usarid: Might this be confusing: "to forbid an instance to have a value" is the same as, or different than, saying the value of the instance is null? Probably best to find different wording.

versions/raml-10/raml-10.md
```diff
@@ -2596,11 +2642,12 @@ The following is an example of various annotation type declarations and the appl
 title: Illustrating annotations
 mediaType: application/json
 annotationTypes:
-  experimental:
-  feedbackRequested:
+  deprecated: null
+  experimental: string | null
```
usarid: This will match "" before it matches `~` so make sure that's what you mean, else reverse the order.

usarid:
See comments, else LGTM

versions/raml-10/raml-10.md
sichvoge: About the fact that `null` is a scalar type; are you sure we want that to be a scalar type? remember that we decided that every scalar type has an additional `enum` facet? that would basically the same for the `null` type, which obviously does not make sense. We could restrict `enum` to be a facet for all except `null`, but do we want that? Why are we actually making `null` a scalar type? I don't really see any reasons. My opinion, it does not have to be a scalar.

usarid: It's a scalar because it has no structure internal to it, unlike a sequence or a map, ie. it doesn't contain anything. You can apply an enum to it too, but the only thing it could have in that enum would be the YAML `~`.

versions/raml-10/raml-10.md
petrochenko-pavel-a: >is the same as, or different than, saying the value of the instance is null?

usually, some languages like javascript have a `undefined` which is different from `null`

>Probably best to find different wording.
Agree. I would say something like null type is a type which has not valid instances.

versions/raml-10/raml-10.md
sichvoge: Thats just a mistake, it needs to be `null | string` of course.

versions/raml-10/raml-10.md
sichvoge: When talking with @petrochenko-pavel-a about that, he explained that it is not necessary to have `null` to be a scalar type vs `null` basically extends `any` (which object, array, and any scalar type extends). So you don't really gain anything, that is what I understood. Therefore, I would not have `null` being a scalar type so that in the future, if necessary, we can reserve to add additional facets for example to scalar w/o necessarily influence the `null` type which should not have any, imho.

sichvoge: OK, as per our meeting today; the best and most consistent approach is to stay with the assumption that `null` is a scalar since it has no structure.

PR #433, opened 2016-04-21
copyedit Inline Type Declarations and Defining Examples in RAML

krishahn:
Copyedit sections:
Inline Type Declarations
Defining Examples in RAML


sichvoge:
LGTM

PR #436, opened 2016-04-22
Resources and nested resources

krishahn:
Copyedit Resources and Nested Resources, no change to already approved paragraph.


sichvoge:
LGTM

PR #437, opened 2016-04-23
copyedit Resource Property thru Template URIs

krishahn:
In Template URIs and URI Parameters, is something missing after "on the wire may look as: /users/%5B%22blue%22,%22green%22%5D/"? Probably just ignorance on my part.


sichvoge:
LGTM

PR #438, opened 2016-04-23
copyedits of Methods, Method-level Display Name, and Method-level des…

krishahn:
Copyedits of Methods, Method-level Display Name, and Method-level Description sections. Does body refer to more than one body?


sichvoge:
LGTM

PR #439, opened 2016-04-24
copyedit Headers section

krishahn:
Copyedit Headers section. This description says, "... or an array type if the underlying type of the array is an object type, _array type_, or a union of non-scalar types...". The Query Parameters in a Query String section says, "... or an array type if the underlying type of the array is an object type or union of non-scalar types..." Is there an unwanted "array type" in the Headers section?


sichvoge:
LGTM

PR #440, opened 2016-04-24
copyedit Query Strings and Query Parameters and The Query String as a…

krishahn:
Copyedit Query Strings and Query Parameters and The Query String as a Whole.


sichvoge:
LGTM

PR #442, opened 2016-04-24
copyedit Method Level Protocols section and Bodies section

krishahn:
copyedit Method Level Protocols section and Bodies section


sichvoge:
LGTM

PR #445, opened 2016-04-25
copyedit Declaration Resource Types and Traits

krishahn:
Copyedit Declaration Resource Types and Traits section


sichvoge:
LGTM

PR #450, opened 2016-04-26
copyedit resource-types-and-traits-effect-on-collections

krishahn:
Copyedit Resource Types and Traits Effect on Collections.


sichvoge:
LGTM

PR #455, opened 2016-04-27
Content adjustmemts for the query parameter section

sichvoge:
According to a short conversation we had, query parameters should treat object types, union types, and arrays in query parameters consistent to headers.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1232,7 +1232,7 @@ types:
 
 In the example above, the type `Teacher` inherits all restrictions from `Person` and `Employee`.
 
-Multiple inheritance is only allowed if the sub-type is still a valid type declaration after inheriting all restrictions from its parent types. For example, the type `Number3` is fully valid:
+Multiple inheritance is allowed only if the sub-type is still a valid type declaration after inheriting all restrictions from its parent types. For example, the type `Number3` is valid:
 
 ```yaml
 types:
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -1227,26 +1227,53 @@ When referencing an inner element of a schema, a RAML processor MUST validate an
...
+    properties:
+      employeeNr: integer
+  Teacher:
+    type: [ Person, Employee ]
 ```
 
-Note: Multiple inheritance is only allowed if all Type Expressions are simple object Types.
+In the example above, the type `Teacher` inherits all restrictions from `Person` and `Employee`.
+
+Multiple inheritance is allowed only if the sub-type is still a valid type declaration after inheriting all restrictions from its parent types. Also, it is not allowed to inherit from different kind of primitive types, for example `[ number, string ]`.
```
usarid: I think you also need to explicitly state how to treat this statement when you have union types somewhere in the inheritance chain. I would guess the answer is to expand all the unions and and see that at least one of the allowed combinations is a valid type declaration and in particular does not combine different primitive types, but @petrochenko-pavel-a should say definitively.

usarid:
One comment, otherwise LGTM

versions/raml-10/raml-10.md
sichvoge: How that works for union types has been mentioned in the union types section already. So I'd would not repeat it here.

sichvoge:
It seems something went wrong with that PR as it includes the changes for the query parameter section and the multiple inheritance chapter. I'll close that and create another one that focuses on only the query parameter changes and merge them as it seems there is no mistake.

Issue #501, opened 2016-05-28
clarify that that the enum facet applies to scalar types only


sichvoge:
The current spec states that "All types, except the file type, can have an additional enum facet." This could potentially create confusion because it's in the section on scalar types, so let's move it outside that section, to a location that clearly applies to all types.

PR #502, opened 2016-05-28
add clarification for the enum facet and that it applies to scalar types only

sichvoge:
According to #501, all types can define the `enum` facet. This PR clarifies that.


versions/raml-10/raml-10.md
```diff
@@ -519,6 +519,7 @@ A type declaration references another type, or wraps or extends another type by
 | (&lt;annotationName&gt;)? | [Annotations](#annotations) to be applied to this API. An annotation is a map having a key that begins with "(" and ends with ")" where the text enclosed in parentheses is the annotation name, and the value is an instance of that annotation.
 | facets? | A map of additional, user-defined restrictions that will be inherited and applied by any extending subtype. See section [User-defined Facets](#user-defined-facets) for more information.
 | xml? | The capability to configure [XML serialization of this type instance](#xml-serialization-of-type-instances).
+| enum? | Enumeration of possible values for this built-in scalar type. The value is an array containing representations of possible values, or a single value if there is only one possible value.
```

>Enumeration of possible values for this built-in scalar type. The value is an array containing representations of possible values, or a single value if there is only one possible value.

to

>An enumeration of all the possible values of instances this type. The value is an array containing representations of these possible values; an instance of this type MUST be equal to one of these values.

Issue #506, opened 2016-05-28
Add a new section for the type `any` 


sichvoge:
All base types such as `object` or the scalar types inherit from a root type called `any`. The current spec should clarify that in a separated section specifically for the type `any`.

Issue #505, opened 2016-05-28
Clarify the built-in types in the specification


sichvoge:
We should create a built-in types section and list all the supported ones in there, and also we need to update the image probably.

PR #507, opened 2016-05-28
introduced built-in types section and clarified the root type any

sichvoge:
that PR clarifies what the root type any is according to #506 and also what built-in types are supported according to #505


versions/raml-10/raml-10.md
```diff
@@ -544,9 +542,31 @@ types:
...
-### Object Types
+### Built-in Types
+
+The RAML type system defines the following built-in types:
+
+* [any](#any-types)
+* [object](#object-types)
+* [array](#array-types)
+* [union](#union-types) via type expression
+* [scalar types](#scalar-types) such as number, boolean, string, date, file, integer, and null
```
usarid: Why "such as"? Is this in fact the definitive list -- in which case, this isn't "such as" but rather an exact list of all scalar types.

versions/raml-10/raml-10.md
```diff
@@ -544,9 +542,31 @@ types:
 
 We recommended using the "type" facet instead of "schema" because the schema alias is deprecated and might be removed in a future RAML version. Also, the "type" facet supports XML and JSON schemas.
 
-### Object Types
+### Built-in Types
```
usarid: I don't quite understand why the section names for the built-in types are pluralized, as if there are many object types, many array types, many "any" types, etc. There is of course only one built-in type for each of these, so the sections should be named accordingly, unless I'm missing something.

versions/raml-10/raml-10.md
```diff
@@ -544,9 +542,31 @@ types:
...
+* [any](#any-types)
+* [object](#object-types)
+* [array](#array-types)
+* [union](#union-types) via type expression
+* [scalar types](#scalar-types) such as number, boolean, string, date, file, integer, and null
+
+Additional to the built-in types, the RAML type system also allows to define [JSON or XML schema](#using-xml-and-json-schema).
 
-All types that have the built-in object type at the root of their inheritance tree can use the following facets in their type declarations:
+#### Any Types
```
usarid: This should be:

#### The "Any" Types

versions/raml-10/raml-10.md
```diff
@@ -544,9 +542,31 @@ types:
...
+
+Every type, whether built-in or user-defined, has the `any` type at the root of its inheritance tree. The "base" type of any type is the type in its inheritance tree that directly extends the `any` type at the root; thus, for example, if a custom type `status` extends the built-in type `integer` which extends the built-in type `number` which extends the `any` type, then the base type of `status` is `number`. Note that a type may have more than one base type.
+
+The `any` type has no facets.
+
+The following diagram shows the inheritance tree, starting at the root-level with `any`.
+
+![Types Hierarchy](images/typesHierarchy.png)
```
usarid: Why is this diagram within the section of the `any` type? It's the general type hierarchy, after all. It should be moved to a section pertaining to all types.

But regardless of where it goes, it's unclear from the diagram whether "scalar" is a type in the type hierarchy from which the various types extend, or it's the name of the branch of the hierarchy. The diagram should be changed to clarify that, especially because it's in a normative section.

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,166 @@ example:
...
+      name: string
+      color: string
+  Dog:
+    type: object
+    properties:
+      name: string
+      fangs: string
+```
+
+A following example of an instance of type `CatOrDog` is valid:
```
usarid: "The following" not "A following"

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,166 @@ example:
 
 Declaring the type of a property to be `null` represents the lack of a value in a type instance. In a RAML context that requires *values* of type `null` (vs just type declarations), the usual YAML `null` is used, e.g. when the type is `null | number` you may use `enum: [ 1, 2, ~ ]` or more explicitly/verbosely `enum: [ 1, 2, !!null "" ]`; in non-inline notation you can just omit the value completely, of course.
 
+#### Union Types
+
+Union types are declared using pipes (|) in type expressions. Union types are useful for modeling common scenarios in JSON-based applications, for example an array containing objects that can be instances of more than one type.
```
usarid: A union type is used to allow instances of data to be described by any of several types. A union type is declared via a type expression that combines 2 or more types delimited by pipe (`|`) symbols; these combined types are referred to as the union type's base types. In the following example, instances of the `Device` type may be described by either the `Phone` type or the `Notebook` type:

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,166 @@ example:
...
+      manufacturer:
+        type: string
+      numberOfUSBPorts:
+        type: number
+      kind: string
+  Device:
+    type: Phone | Notebook
+```
+
+To deserialize an instance of data, first fully expand all unions at every level, then match the instance against each element in that expansion starting from the left-most and proceeding to the right; use the first successful matching element to deserialize the instance; and if no elements match, the instance is invalid. A valid instance of a union type must pass all restrictions associated with at least one of the elements. For example:
```
usarid: Change to:

An instance of a union type is valid if and only if it meets all restrictions associated with at least one of the base types. More generally, an instance of a type that has a union type in its type hierarchy is valid if and only if it is a valid instance of at least one of the base types obtained by expanding all unions in that type hierarchy. Such an instance is deserialized by performing this expansion and then matching the instance against all the base types, starting from the left-most and proceeding to the right; the first successfully-matching base type is used to deserialize the instance.

The following example defines two types and a third type which is a union of those two types.

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,166 @@ example:
...
+/person:
+  get:
+    responses:
+      200:
+        body:
+          application/json:
+            type: !include person.json
+```
+
+A RAML processor MUST not allow types that define an XML or JSON schema to participate in type inheritance or specialization, or effectively in any [type expression](#type-expressions). Therefore, you cannot define sub-types of these types to declare new properties, add restrictions, set facets, or declare facets. You can, however, create simple type wrappers that add annotations, examples, displayName, or a description.
```
usarid: "MUST not" --> "MUST NOT"

usarid: displayName --> display name

versions/raml-10/raml-10.md
```diff
@@ -1932,7 +1952,7 @@ An API method can support or require a query string in the URL on which the meth
 
 #### The Query String as a Whole
 
-The **queryString** node is used to specify the query string as a whole, rather than as name-value pairs. The queryString value is either the name of a data type or an inline data type declaration, including a data type expression. In either case, all types at the root of the type hierarchy of the data type MUST be either a scalar type or the object type, after fully expanding any union type expressions at every level of the type hierarchy.
+The **queryString** node is used to specify the query string as a whole, rather than as name-value pairs. The queryString value is either the name of a data type or an inline data type declaration, including a data type expression. In either case, all types at the base of the type hierarchy of the data type MUST be either a scalar type or the object type, after fully expanding any union type expressions at every level of the type hierarchy.
```
usarid: "all types at the base of the type hierarchy of the data type" --> "all base types in type hierarchy of the data type"

versions/raml-10/raml-10.md
sichvoge: Guess that's obsolete since we describe the build-in types now. Agree!

sichvoge:
On left:

- [ ] Update the type hierarchy image and put it into the respective chapter

Currently the image is not in this version; need to work on that a bit more and put it into its right place.

versions/raml-10/raml-10.md
```diff
@@ -544,9 +542,31 @@ types:
...
+Additional to the built-in types, the RAML type system also allows to define [JSON or XML schema](#using-xml-and-json-schema).
+
+The following diagram shows the inheritance tree, starting at the root-level with `any`.
+
+![Types Hierarchy](images/typesHierarchy.png)
+
+#### The "Any" Type
 
-All types that have the built-in object type at the root of their inheritance tree can use the following facets in their type declarations:
+Every type, whether built-in or user-defined, has the `any` type at the root of its inheritance tree. The "base" type of any type is the type in its inheritance tree that directly extends the `any` type at the root; thus, for example, if a custom type `status` extends the built-in type `integer` which extends the built-in type `number` which extends the `any` type, then the base type of `status` is `number`. Note that a type may have more than one base type.
```
usarid: We've forgotten the most important part: what *is* the `any` type? It's a type that imposes no restrictions, i.e. any instance of data is valid against it. Maybe get more strict language from @petrochenko-pavel-a

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,168 @@ example:
 
 Declaring the type of a property to be `null` represents the lack of a value in a type instance. In a RAML context that requires *values* of type `null` (vs just type declarations), the usual YAML `null` is used, e.g. when the type is `null | number` you may use `enum: [ 1, 2, ~ ]` or more explicitly/verbosely `enum: [ 1, 2, !!null "" ]`; in non-inline notation you can just omit the value completely, of course.
 
+#### Union Type
+
+A union type is used to allow instances of data to be described by any of several types. A union type is declared via a type expression that combines 2 or more types delimited by pipe (`|`) symbols; these combined types are referred to as the union type's base types. In the following example, instances of the `Device` type may be described by either the `Phone` type or the `Notebook` type:
```
usarid: I just noticed we use "base" types here differently than how we used them in the `any` type definition and in the various places where we talk about base types. We should change the term in one of these to not confuse it with the other.

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,168 @@ example:
...
+
+```yaml
```
usarid: Shouldn't all our examples of RAML use the raml language, not yaml?

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,168 @@ example:
...
+    properties:
+      name: string
+      fangs: string
+```
+
+The following example of an instance of type `CatOrDog` is valid:
+
+```yaml
+CatOrDog: # follows restrictions applied to the type 'Cat'
+  name: "Musia",
```
usarid: No need for quotes here or in the next line

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,168 @@ example:
...
+       name: string
+       fangs: string       
+   HomeAnimal: [ HasHome ,  Dog | Cat ]
+```
+
+In this case, type `HomeAnimal` has two base types, `HasHome` and an anonymous union type, defined by the following type expression: `Dog | Cat`.
+
+Validating the `HomeAnimal` type involves validating the types derived from each of the base types and the types of each element in the union type. In this particular case, you need to test that types `[HasHome, Dog]` and `[HasHome, Cat]` are valid types.
+
+If you are extending from two union types you should perform validations for every possible combination. For example, to validate the `HomeAnimal` type shown below, you need to test six possible combinations: `[HasHome, Dog ]`, `[HasHome, Cat ]`, `[HasHome, Parrot]`, `[IsOnFarm, Dog ]`, `[IsOnFarm, Cat ]`, and `[IsOnFarm, Parrot]`.
```
usarid: Why "should" vs "must"? This should all be normative and clear-cut.

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,168 @@ example:
...
+/person:
+  get:
+    responses:
+      200:
+        body:
+          application/json:
+            type: !include person.json
+```
+
+A RAML processor MUST NOT allow types that define an XML or JSON schema to participate in type inheritance or specialization, or effectively in any [type expression](#type-expressions). Therefore, you cannot define sub-types of these types to declare new properties, add restrictions, set facets, or declare facets. You can, however, create simple type wrappers that add annotations, examples, display name, or a description.
```
usarid: Type wrappers are inheritance, so we need to restate that sentence: basically there are restrictions on the type inheritance that XML and JSON schema can participate in. Let's not talk about "specialization" unless we've defined that somewhere.

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,168 @@ example:
...
+  Board:
+    properties:
+      members: Person[] # invalid use of type expression '[]' and as a property type
+```
+
+A RAML processor MUST be able to interpret and apply JSON Schema and XML Schema.
+
+An XML schema, or JSON schema, MUST NOT be used where the media type does not allow XML-formatted data, or JSON-formatted data, respectively. XML and JSON schemas are also forbidden in any declaration of query parameters, query string, URI parameters, and headers.
+
+The nodes "schemas" and "types", as well as "schema" and "type", are mutually exclusive and synonymous for compatibility with RAML 0.8. API definitions should use "types" and "type", as "schemas" and "schema" are deprecated and might be removed in a future RAML version.
```
usarid: Need more normative language

versions/raml-10/raml-10.md
```diff
@@ -1070,6 +1090,168 @@ example:
...
+
+##### References to Inner Elements
+
+Sometimes it is necessary to refer to an element defined in a schema. RAML supports that by using URL fragments as shown in the example below.
+
+```yaml
+type: !include elements.xsd#Foo
+```
+
+When referencing an inner element of a schema, a RAML processor MUST validate an instance against that particular element. This version of RAML specification supports referencing any inner elements in JSON schemas that are valid schemas, any globally defined elements, and complex types in XML schemas. There are only a few restrictions:
```
usarid: Why talk about "this version" when we don't talk that way elsewhere in the spec? The first restriction below hardly seems like a restriction, so I'm not sure what is the point here. The second bullet point also does not seem like a restriction, more like an observation.

versions/raml-10/raml-10.md
```diff
@@ -3230,7 +3252,7 @@ resourceTypes:
...
+In the document that applies a library, the data types, resource types, traits, security schemes, and annotation types in the library become available from the document. The assets in the library are referenced from the document using dot notation as follows: concatenate the library name followed by a period ("."), and then the name of the data type, resource type, trait, security scheme, or annotation type. The library name defines a namespace for the library nodes within the context that the library is applied. Namespaces defined in a **uses** statement in a specific file are only consumable within that file and serve only to disambiguate the included libraries from each other. Therefore, any processor MUST NOT allow any composition of namespaces using "." across multiple libraries.
```
usarid: "In the document..." --> "In a document...", and "from the document" --> "within the document" in a couple of places. Also: "and then the name of the data type" --> "followed by the name of the data type". "within the context that the library" --> "within the context in which the library".

usarid:
See comments inline.

versions/raml-10/raml-10.md
sichvoge: its effectively the same, but YAML is more commonly supported by markdown interpreters. Hence, we will not get any problems later.

versions/raml-10/raml-10.md
petrochenko-pavel-a: I would like only add that  `any` type has no `facets` and suggest to insert a following wording:

```
By definition `any` type is type which has no restrictions and no facets associated with it. So every instance of data is valid against it.
```
after

>has the `any` type at the root of its inheritance tree.

versions/raml-10/raml-10.md
```diff
@@ -544,9 +542,31 @@ types:
...
+#### The "Any" Type
 
-All types that have the built-in object type at the root of their inheritance tree can use the following facets in their type declarations:
+Every type, whether built-in or user-defined, has the `any` type at the root of its inheritance tree. The "base" type of any type is the type in its inheritance tree that directly extends the `any` type at the root; thus, for example, if a custom type `status` extends the built-in type `integer` which extends the built-in type `number` which extends the `any` type, then the base type of `status` is `number`. Note that a type may have more than one base type.
+
+The `any` type has no facets.
+
+#### Object Type
+
+All types that have the built-in object base type can use the following facets in their type declarations:
```
petrochenko-pavel-a: The language here looks a bit strange: `that have the built-in object base type`, I would probably prefer to say some thing like  `that have the built-in object base type in their super types hierarchy`

versions/raml-10/raml-10.md
petrochenko-pavel-a: I would prefer to use wording `option types` here. Also it seems that we need to add a line saying that you may effectively consider narrowest common super types of option types a super types of the union type.

sichvoge:
@usarid I have created a couple of separated issues for some of your suggested improvements to the content. This PR should concentrate on the any type clarification only. See my latest commit for the any type enhancement to the respective chapters

usarid:
LGTM assuming that my comments from 26 days ago will be addresses in separate PRs

Issue #513, opened 2016-06-21
!lowercamelcase example probably should not be userId --> userId


hillazz:
Hi!
Probably the example should be UserId --> userId but not userId --> userId. This can be found in "Resource Type and Trait Parameters" chapter.

Thanks!

sichvoge:
Happy to accept any PR ;)

hillazz:
Hi, here is the PR: https://github.com/raml-org/raml-spec/pull/519

sichvoge:
Thanks!

PR #519, opened 2016-06-29
#513 fix - updated !lowercamelcase modifier example to UserId --> userId

hillazz:
RAML 1.0 - updated !lowercamelcase modifier example from userId --> userId to UserId --> userId. It looks more clear.


sichvoge:
LGTM 👍  Thanks @hillazz. That fixes #513

PR #529, opened 2016-07-08
Trait merging

ddenisenko:
Refactored trait merging algorithm structure. Modified merging algorithm to use YAML terms. Added parameters application and reference patching sections.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -107,8 +107,7 @@ Throughout this specification, **Markdown** means [GitHub-Flavored Markdown](htt
 	- [Applying Resource Types and Traits](#applying-resource-types-and-traits)
 	- [Resource Type and Trait Parameters](#resource-type-and-trait-parameters)
 	- [Declaring HTTP Methods as Optional](#declaring-http-methods-as-optional)
-	- [Algorithm of Merging Traits and Methods](#algorithm-of-merging-traits-and-methods)
-	- [Resource Types and Traits Effect on Collections](#resource-types-and-traits-effect-on-collections)
+	- [Algorithm of Merging Traits and Resource Types](#algorithm-of-merging-traits-and-resource-types)
 - [Security Schemes](#security-schemes)
 	- [Security Scheme Types](#security-scheme-types)
 	- [Security Scheme Declaration](#security-scheme-declaration)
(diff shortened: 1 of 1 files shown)


ddenisenko:
@sichvoge please review and modify if needed

versions/raml-10/raml-10.md
```diff
@@ -2367,18 +2366,86 @@ resourceTypes:
...
+  secured:
+    queryParameters:
+      <<tokenName>>:
+        description: A valid <<tokenName>> is required
+/servers:
+  get:
+    is: [ { secured : { tokenName: token } } ]
+```
+
+To resolve a collision arising from this inequality, priority is given to the trait in closest proximity to the target method or resource. In the previous example, the `tokenName` parameter value for the `GET:/servers` method is `token`, and the trait list consists of single trait occurrence: `[ {secured:{ tokenName:token}} ]`.
```
KonstantinSviridov: Actually, we support applying same trate with different parameters https://github.com/raml-org/raml-js-parser-2/issues/94

versions/raml-10/raml-10.md
```diff
@@ -2433,54 +2502,216 @@ Every explicit node wins over the ones that are declared in a resource type or t
...
+
+##### Reference resolving
+
+For each reference a queue of RAML units (files) is calculated and an algorithm should look for the reference target in each unit, in turn. The first matching reference name and kind is treated as a result. For the example above, we perform the resolving of each `Monkey` reference in Phase 2 results.
+
+###### Static reference resolving
+
+In case of reference value being NOT a result of template parameter substitution, the units queue is opposite to the units and order of node traversing in Phase 1 leading from initial resource or method node to the reference application node.
+
+For the example above and `hA2: Monkey` reference phase 1 algorithm starts from `/users` resource node defined in unit `main.raml` and ends in `hA2: Monkey` node in `libA.raml` unit. So the final unit sequence for phase 3 is opposite to that order: [`libA.raml`,`main.raml`].
```
svacas: a static reference should only be checked in the context where it was defined

versions/raml-10/raml-10.md
```diff
@@ -2433,54 +2502,216 @@ Every explicit node wins over the ones that are declared in a resource type or t
...
+
+As a result in this case, `kind: Monkey` will be resolved to `Monkey` type defined in `main.raml` as this unit is the first one in queue containing type called `Monkey`.
+
+For the example above and `hB: <<kindB>>` reference defined in `libB.raml`, the node setting its value is `kindB: <<kind>>` in `libA.raml`, and its value is in turn set by `kind: Monkey` in `main.raml`. So the unit sequence for phase 3 is phase 1 path from `kind: Monkey` (`/users`) to `hB: <<kindB>>`, and that is [`main.raml`, `libA.raml`, `libB.raml`].
+
+As a result in this case, `kind: Monkey` will be resolved to `Monkey` type defined in `main.raml` as this unit is the first one in queue containing type called `Monkey`.
+
+##### Reference patching
+
+After reference target is resolved, reference text must be patched to become a correct reference from the main unit to the previously found reference target.
```
svacas: this patching mechanism looks like an implementation detail and shouldn't be required as long as the references are correctly resolved.

versions/raml-10/raml-10.md
```diff
@@ -2367,18 +2366,86 @@ resourceTypes:
...
 
-1. Method node properties are inspected and those that are undefined in trait node remain unchanged.  
-2. The method node receives all properties of trait node (excluding optional ones), which are undefined in the method node.  
-3. Properties defined in both method node and trait node (including optional ones) are treated as follows:  
-  * Scalar properties remain unchanged.  
-  * Collection properties are merged by value, as described later.  
-  * Values of object properties are subjected to steps 1-3 of this procedure.  
+Algorithm has three phases:
+1. YAML-level merging
+2. Parameters application
```
svacas: parameters should be applied before merging, otherwise key parameters will not be correctly merged

versions/raml-10/raml-10.md
ddenisenko: Well, these patched references are going into JSON. Users are going to see them. Why do you think its implementation details?

versions/raml-10/raml-10.md
ddenisenko: Good catch! In my tries to simplify the description I went too far. In practise, imho phases 1 and 2 should be performed simultaneously.

Applying parameters before merging, while possible, is probably going to complicate algorithm even more: the same trait / resource type can be used along the hierarchy with different parameters, so we'll have to save a copy for a trait/resource type and all of its dependencies (or at least the ones affected by parameter value) for each set of parameter values. 

I'll try to refactor this description for a single pass of phases 1 and 2.

versions/raml-10/raml-10.md
svacas: AFAIK the JSON serialization is not part of the spec, otherwise it should be there.

ddenisenko: Yes, I agree, JSON is not a part of the spec, so theoretically we may probably skip specifying how the result will look like. 

But, taking into account we're trying to make parsers compatible and this JSON is something common between the parsers, why not to make clear what exactly to do with the references?

svacas: the problem I see is that in this section we provide information related to the json output, but we are not doing the same in all the other chapters. For me everything concerning the json output should be added to an appendix that describes the TCK serialization and is optional to implement.

versions/raml-10/raml-10.md
svacas: maybe fragments should not define their own context unless they have libraries themselves, and act as regular non-fragment includes. If they have libraries then they define their own context and cannot access references from the parent context.
If we want fragments defining their own context always (no matter if they use libraries or not) then the user will need to use a non-fragment include in order to use the references from the parent file.

versions/raml-10/raml-10.md
antoniogarrote: Hi, I agree with @svacas about the patching mechanism. You shouldn't need it if you have already resolved references.
But I think the rules for visibility, scopes, etc are not very well defined. I can only find the [Applying Libraries](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#applying-libraries) section in the spec.

Some questions about it:

- Does every trait file open a new namespace and declaration scope?
- Does these library namespaces have a name or are 'anonymous' namespaces?
- I think it confuses an alias for a namespace (what we use to reference the library in the document) with the actual identifier of the namespace (that I think right now is anonymous)
- Does the same rules for `includes` apply to `uses` statement to locate a library? Shouldn't they be all of the library references treated just as URIs?

For example, is this:

```yaml
#%RAML 1.0
# libA.raml

types:
  Monkey: string
```

actually:

```yaml
#%RAML 1.0
# file://libA.raml

types:
  file://libA#Monkey: string
```
And if that's the case, doing this:

```yaml
#%RAML 1.0
#% main.raml

uses:
  a: libA.raml

/users:
  post:
    headers:
      hA2: a.Monkey
```
Just a shortcut for:

```yaml
#%RAML 1.0
#% file://main.raml

/users:
  post:
    headers:
      hA2: file://libA#Monkey
```

- If that's the case, is there a way to reference a symbol directly without a `uses` statement

Something like:

```RAML
#%RAML 1.0
#% main.raml

/users:
  post:
    headers:
      hA2: http://raml.org/traits/libA#Monkey
```
It looks like RAML has a clear notion of modularity but lacks a well defined way of referencing the information in the modules.

versions/raml-10/raml-10.md
ddenisenko: The concept of an included file either defining or not defining its own context basing on some circumstances  will probably cause a lot of confusion for the users.

Basing functionality on a split of includes into "fragments" and "not fragments" will probably also make it hard.

I'd suggest one of the two approaches: we either allow resolving static references from all the contexts along the line, or we only allow current context and make an official exception in case of includes.

What do you think?

versions/raml-10/raml-10.md
ddenisenko: These questions are more about the whole reference resolving system rather than about traits and resource types.

>Does every trait file open a new namespace and declaration scope?
Every RAML file opens a new namespace and declaration scope. Fragments and non-fragment includes are in question at the moment.

>Does these library namespaces have a name or are 'anonymous' namespaces?
Their are anonymous. There is not global library identifier/URL.

>I think it confuses an alias for a namespace (what we use to reference the library in the document) with the actual identifier of the namespace (that I think right now is anonymous)

There is not confusion in this case because there is no such thing as the actual identifier of the namespace/unit in the spec. The only way of library identification is the path in uses instruction and thus identification is always context-dependent.

>Does the same rules for includes apply to uses statement to locate a library? Shouldn't they be all of the library references treated just as URIs?

`#` notion in include paths does not work for RAML files if I am not missing something. This one is for schemas. uses instruction does not support `#` at all.

There is absolutely no way of referencing something directly without an alias set up by `uses` instruction, and that is the reason why I want the reference patching to be the part of the spec.

See, the whole trait merging algorithm is the algorithm of translation of one RAML tree (or RAML trees) into another one. The result of such translation should presumably be a valid RAML. While it is not mentioned in the spec, that result should be printed at some point of time, whether it will be JSON or some other form of output.

If I got it right, you want to cut off the last part of this translation. So, we do describe how exactly we modify the resulting tree by adding contents of a trait to a method, we describe how we modify the resulting tree by applying parameters, but we want to stop here and be silent regarding how we modify the tree to make the references valid/resolvable? Just mention that "references should be valid in the result" and that's it? Why so?

sichvoge:
@ddenisenko where are we with this?

PR #558, opened 2016-08-05
resourcePath and resourcePathName are in the wrong order.

redndahead:
It says "sets the value of resourcePathName and resourcePath parameters to "/bom/{itemId}" and "bom", respectively."
Those should actually be reversed.


sichvoge:
I think reading the spec, it seems that both `resourcePathName` and `resourcePath` will result into `"bom"`. @petrochenko-pavel-a can you help to clarify?

PR #574, opened 2016-09-22
Proposed clarification for wrapped.

petrochenko-pavel-a:
Suggestion for inner xml element name clarification. I am not sure about naming for inner elements in the case of object types. (primary use case for wrapped + object types is solving polymorphism which is loosely specified in the spec)


sichvoge:
Can you work a little bit more on the text @petrochenko-pavel-a. I am not sure if it is completely clear. We can also add another example, or two, to clarify the naming convention for wrapped nodes.

Issue #578, opened 2016-09-28
Clarify behavior of consecutive slashes due to endpoints with trailing slash.


tomchristie:
This is related to #256, but applies to consecutive slashes created between resources.

If an API has a convention of using trailing slashes on resources, then it's unclear how RAML expects that to be treated.

For example, suppose our endpoints are:

    /users/           # Note the trailing slash
    /users/{pk}/   # Note the trailing slash

Presumably that should that be represented as:

    /users/:
        get: ...
        /{pk}/:
           get: ...

However clients might well end up treating that as `/users//{pk}/`, rather than removing the extraneous slash, which could be problematic.

sichvoge:
As in #256, those trailing slashes are omitted. What's the use case having trailing slashes or even `/users//{pk}/`?

tomchristie:
> What's the use case having trailing slashes

Trailing slashes for API endpoints is a valid enough style, just a different convention to *not* having trailing slashes. What I'd like to make sure is that I can specify a RAML document for an API that uses this style of convention:
 
    /users/
    /users/{pk}/

As well as being able to specify a RAML document for an API that uses this style of convention

    /users
    /users/{pk}

Right now it's not totally clear how you're supposed to express the former convention in RAML, because by the spec it that both the trailing and leading slashes would be concatenated.

tomchristie:
In fact, the spec *explicitly* doesn't allow for this convention, given https://github.com/raml-org/raml-spec/pull/275/files#diff-5a7129c2f7b175872e69eb61ea801326R237

tomchristie:
Couple of posts on the APIs "trailing slash" or "no-trailing slash" convention, for context:

http://apievangelist.com/2015/08/04/how-do-we-handle-that-trailing-slash-in-our-api-endpoints/
https://cdivilly.wordpress.com/2014/03/11/why-trailing-slashes-on-uris-are-important/

sichvoge:
The question is if they are really different resources? I am not seeing a single opinion on that :)

tomchristie:
> The question is if they are really different resources?

The API spec needs to be able to correctly describe whichever is the canonical URL.

A framework *might* happen to handle both `/users` and `/users/` as exactly the same, or more typically it will redirect from the non-canonical case to the canonical case. (Which means we'd always end up making duplicated requests for GET requests, and may not even be supported by clients for non-GET requests)

tomchristie:
[...]

PR #580, opened 2016-10-06
Refine trailing slash behavior.

tomchristie:
This proposal does two things:

* Modify the wording of how trailing slashes are treated, in order to allow for resources that (1) have a trailing slash and (2) also have a child resource.
* Changes the example to a more conventional example case. (the existing example with multiple consecutive slashes within the URL did not present a typical real-world case, and is more confusing than helpful.)

Without this change it is impossible to represent the `http://api.test.com/common/users/` and `http://api.test.com/common/users/groups/` resources as desired, because the absolute path resolution for the groups resource would result in `http://api.test.com/common/users//groups/`.

If required I could present this pull request *without* also cleaning up the example to be more conventional, but I think it would be less clear.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -317,13 +317,12 @@ baseUri: http://api.test.com/common/
   /groups:
 ```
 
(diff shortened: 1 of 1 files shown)


tomchristie:
Refs: #578

PR #599, opened 2017-01-28
Add baseUri to example

zacyang:
In section https://github.com/raml-org/raml-spec/blob/master/versions/raml-08/raml-08.md#template-uris-and-uri-parameters, there are two example missing the required key baseUri

I understand it claims in the previous section that baseUri is 
> Optional during development; Required after implementation

As I am building a parser for raml and extracting test case according to this spec and found this example is not able to pass the previous validation on baseUri because it is missing.
Personally, I think it would be nice to have the baseUri section added to those two examples.


sichvoge:
Might be missing something, but the `baseUri` is not mandatory.

PR #634, opened 2017-05-31
fixed various forms of typos

GeraldLoeffler:
These fixes should be uncontroversial hence i'm sending them as a PR rather than filing a bug.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -700,13 +700,14 @@ types:
         type: string
 ```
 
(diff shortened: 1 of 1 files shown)


sichvoge:
Hi @GeraldLoeffler. Thanks for the PR. I have created a separated PR to consolidate several other fixes and brought in some of your changes. Others I haven't brought in since I think you misunderstood something :)

For example, there is a difference between `comment?: string` and `comment: string?`. The former is an optional property and the latter is more about missing values and `nil` which can also be expressed as `comment: nil | string`.

PR #653, opened 2017-10-02
Fix Projects link in README.md

boliza:
fix `Projects` broken link


sichvoge:
LGMT! Good catch btw ;)

PR #656, opened 2017-10-15
Update pattern property example

tomasliubinas:
In the current example of pattern property specification `note1` property is described as valid as not matching the pattern. The property is valid however for the different reason. It is valid as it is a string. This pull request provides a correction for the comment and adds an example of another property not matching the pattern.


jstoiko:
Long overdue reply! Thanks for the PR @tomasliubinas. Merging.

jstoiko (approved):

PR #689, opened 2018-08-21
Since raml-parser is using SnakeYAML, it supports YAML 1.1 

asomov:
The support for YAML 1.2 is implemented in snakeyaml-engine (which is now under construction)


jstoiko:
@asomov: I suppose that by `raml-parser`, you are referring to `raml-java-parser`. In any case, the version  of YAML that a particular parser supports (whether it be Java, JavaScript, Golang, PHP, etc) is irrelevant. The RAML 1.0 Spec is based on YAML 1.2 and we are not going to change that.

PR #698, opened 2018-11-01
[0.8] Fix yaml examples, refs #697

MikeRalphson:
* Two missing `:` characters on YAML keys
* Two missing spaces between `:` and key value


jstoiko:
Nice to see you here Mike. Thanks for this!

MikeRalphson:
> Nice to see you here Mike. Thanks for this!

No problem (it was me publishing a new version of `mdv` which caused the build failure / the incorrect examples to be spotted...)

PR #700, opened 2018-11-21
Typo in the `Object Type Specialization` sub-section

Sergiu-Ghita:


Sergiu-Ghita:
Thanks @jstoiko, you are correct

PR #732, opened 2020-08-15
Clarify use of question marks in key names

### LINKED ISSUE #707: Clarify use of trailing question mark
In the ["General" sub-section](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#general) of the [Definitions and Terminology section](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#definitions-and-terminology), this statement needs to be clarified:

> A trailing question mark, for example description?, indicates an optional property.

### Proposed clarification:
- remove `, for example description?`
- change sentence to `A trailing question mark MAY indicate an optional property.`
- explicitly mention which properties, in each section, may use such trailing question mark and why, E.g. by adding a sentence starting with `A trailing question mark may be used here to ...`:
  - [`Property Declarations`](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#property-declarations) of an object type
  - [`Declaring HTTP Methods as Optional`](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#declaring-http-methods-as-optional)

jstoiko:
Closes #707

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -598,7 +598,7 @@ types:
 
 Properties of object types are defined using the OPTIONAL **properties** facet. The RAML Specification calls the value of the "properties" facet a "properties declaration". The properties declaration is a map of keys and values. The keys are valid property names for declaring a type instance. The values are either a name of a type or an inline type declaration.
 
-The properties declaration can specify whether a property is required or optional. Alternatively, a trailing question mark (`?`) in the key name MAY be used to indicate that said property is optional.
+The properties declaration can specify whether a property is required or optional. Alternatively, a trailing question mark (`?`) in the key name MAY be used to indicate that a property is optional.
 
 | Facet  | Description |
 |:----------|:----------|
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -1992,7 +1992,7 @@ types:
...
+The **queryParameters** node specifies the set of query parameters from which the query string is composed. When applying the restrictions defined by the API, processors MUST regard the query string as a set of query parameters according to the URL encoding format. The value of the queryParameters node is a [properties declaration](#property-declarations) object, as is the value of the properties object of a type declaration. Each property in this declaration object is referred to as a **query parameter declaration**. Each property name specifies an allowed query parameter name. A trailing question mark (`?`) in a property name MAY be used to indicate that said query parameter is optional. Each property value specifies the query parameter value type as the name of a type or an inline type declaration.
```
dukesphere, line 1995:
```suggestion
The **queryParameters** node specifies the set of query parameters from which the query string is composed. When applying the restrictions defined by the API, processors MUST regard the query string as a set of query parameters according to the URL encoding format. The value of the queryParameters node is a [properties declaration](#property-declarations) object, which is also the value of the properties object of a type declaration. Each property in this declaration object is referred to as a **query parameter declaration**. Each property name specifies an allowed query parameter name. A trailing question mark (`?`) in a property name MAY be used to indicate that a query parameter is optional. Each property value specifies the query parameter value type as the name of a type or an inline type declaration.
```

dukesphere (commented):
minor tweaks

dukesphere (approved):

Issue #665, opened 2017-11-29
Nullable type


carowright:
There is an example in nil type section that adds a question mark on the property type side: 

````
types:
  NilValue:
    type: object
    properties:
      name:
      comment: nil | string # equivalent to ->
                             # comment: string?
    example:
      name: Fred
      comment: # Providing a value or not providing a value here is allowed.
````

As it's mentioned in the comment, it's the same as:

````
types:
  NilValue:
    type: object
    properties:
      name:
      comment: string?
    
example:
      name: Fred
      comment: # Providing a value or not providing a value here is allowed.
````

This example can be found in [here](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#nil-type).
Besides this example there is no mention of how this construction works. So, either the example is invalid or this construction needs to be defined somewhere in the spec.

**Comment by sichvoge** [comment]
Hi Caro. It is defined, but not very clear indeed. Look into the first example of yours. the important line is:

```yaml
comment: nil | string # equivalent to ->
                      # comment: string?
```

As you can see, `comment: string?` is syntactic sugar and equivalent to `comment: nil | string`. 

I am sure we could clarify this a bit more. If you have time, may I ask you to send us a PR?

alvassin:
@sichvoge seems to be a bug.

according to `nil | string` equals to `comment?` means that `comment` is optional, but it should be not like that.

technically comment must be required, but can have as null as string values as well.

jstoiko:
@alvassin: `comment?` and `string?` are two different things.

```raml
comment: nil | string
```
is equivalent to:
```raml
comment: string?
```
and means that the value of `comment` can be `null` whereas:
```raml
comment?: string
```
means that `comment` is not required.

We could also have:
```raml
comment?: string?
```
which would mean that `comment` can be `null` and is also not required.

sheamunion:
@jstoiko I'm trying to create a required, nullable property with a user-defined type in a separate file. For some reason, my IDE (MuleSoft Anypoint Studio) considers this invalid. Am I trying to do something invalid or is my IDE incorrect?

**Response.raml**
```
#%RAML 1.0 DataType

type: object
properties:
  error: nil | !include Error.raml
```

**Error.raml**
```
#%RAML 1.0 DataType

type: object
properties:
  code: string
  message: string
```

jstoiko:
[...]

Issue #569, opened 2016-09-08
Nil and required differences could be ambiguous and hard to understand


boillodmanuel:
Hello,

Today, RAML (And json schema too) distinguish two concepts:
- required: should the property be present in the payload, whatever the value
- nil: could the property value be null

These two concepts are a bit technical and for end-users, the question is often limited to "should I set a value or not". 

The difference between **property missing** or **property present with null value** is often a serialization / technical problem.

I've been working with Swagger and RAML for a long time now and always had to read the doc several times to be sure of my understanding, and actually, I'm never sure :) When I asked my team's members, it turned out this was confusing for them too.

So, I will try to summarize what I understand here.

# Required value

In my examples, I will use the following `Person` type and focus on `surname` attribute:

```yaml
types:
  Person:
    properties:
      name:
        type: string
      surname:
        type: string
```

With this definition, the description property should be present and its value should not be null, ie:

- `{ "name": "Mickael",  "surname": "Mike" }` is VALID
- `{ "name": "Mickael",  "surname": null }` is NOT valid
- `{ "name": "Mickael"}` is NOT valid

# Optional value

If a user cannot fill the `surname` field (in a form for example), we should use this RAML definition:

```yaml
types:
  Person:
    properties:
      name:
        type: string
      surname?:
        type: string?
```

or the more verbose one:

```yaml
types:
  Person:
    properties:
      name:
        type: string
      surname:
        type: string | nil
        required: false
```

In this case, we have:
- `{ "name": "Mickael",  "surname": "Mike" }` is VALID
- `{ "name": "Mickael",  "surname": null }` is VALID
- `{ "name": "Mickael"}` is VALID

IMHO when we are talking about an optional property (at business level), this is what we have in mind.

So there are two interesting/common cases: 
1. `required and type not containing nil`
1. `not required or type containing nil`. 

Most of the people I talk about this issue with think `required` attribute means option 1 and not  option 2 and don't ever use type: nil anywhere. They don't consider the remaining options (required but can be nil, not required but cannot be nil) because not specifying a property is most of the time treated as if the result was there but nil. The difference only is in the serializer's configuration.

I wonder if there is a real demand/need to distinguish other options?

By the way, I think  that developers should use the `required: false` and `type: string | nil` to define optional property (always at business level). I don't thing the developers are not doing this, may be due to bad understanding of the spec. Am I right?

If true, I think it might be a good idea to remove this ambiguity in the documentation.
What do you think?

# Optional request body

[...]

PR #733, opened 2020-08-17
Clarify nilable type syntax regarding unions

### LINKED ISSUE #708: Clarify nilable property alternative syntax
The only place where this alternative syntax of nilable property is mentioned is in a comment on the last example of the [Nil Type](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#nil-type). See also #665.

### Proposed clarification
- at the end of the first paragraph of the ["Nil Type"](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#nil-type) section, add a sentence about "nilable". e.g. 

> Using the type `nil` in a union of one or multiple types, makes a type definition nilable, meaning that its value is allowed to be `null`. When such union is composed of only one type in addition to `| nil`, a trailing question mark `?` in lieu of such union is equivalent. 

- remove the `equivalent to comment: string?` from that example
- add a sentence at the end of that section explaining the equivalent syntax along with an example
- make sure this syntax can only be applied to scalar types and type references, i.e. forbidden in type expressions
- address corner-case of user-defined type with trailing question mark (see comment below)

jstoiko:
Closes #708

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1036,7 +1036,7 @@ types:
 
 In RAML, the type `nil` is a scalar type that allows only nil data values. Specifically, in YAML it allows only YAML's `null` (or its equivalent representations, such as `~`), in JSON it allows only JSON's `null`, and in XML it allows only XML's `xsi:nil`. In headers, URI parameters, and query parameters, the `nil` type only allows the string value "nil" (case-sensitive); and in turn an instance having the string value "nil" (case-sensitive), when described with the `nil` type, deserializes to a nil value.
 
-Using the type `nil` in a union, makes a type definition nilable, meaning that any instance of that union MAY be a nil data value. When such union is composed of only one type in addition to `| nil`, a trailing question mark `?` in lieu of such union is equivalent. The use of that equivalent alternative syntax SHALL be restricted to [scalar types](#scalar-types) and references to user-defined types, and SHALL NOT be used in [type expressions](#type-expressions).
+Using the type `nil` in a union makes a type definition nilable, which means that any instance of that union MAY be a nil data value. When such a union is composed of only one type in addition to `| nil`, use of a trailing question mark `?` instead of the union syntax is equivalent. The use of that equivalent, alternative syntax SHALL be restricted to [scalar types](#scalar-types) and references to user-defined types, and SHALL NOT be used in [type expressions](#type-expressions).
 
 In the following example, the type is an object and has two required properties, `name` and `comment`, both defaulting to type `string`. In `example`, `name` is assigned a string value, but comment is nil and this is _not_ allowed because RAML expects a string.
 
@@ -1083,7 +1083,7 @@ types:
 
 Declaring the type of a property to be `nil` represents the lack of a value in a type instance. In a RAML context that requires *values* of type `nil` (vs just type declarations), the usual YAML `null` is used, e.g. when the type is `nil | number` you may use `enum: [ 1, 2, ~ ]` or more explicitly/verbosely `enum: [ 1, 2, !!null "" ]`; in non-inline notation you can just omit the value completely, of course.
 
-Nilable values are not the same as optional properties, e.g. to have a `comment` property that is optional and which can also have a `nil` value, you could define it as `comment?: string?` or `comment?: nil | string`.
+Nilable values are not the same as optional properties. For example, you can define a `comment` property that is optional and that accepts a `nil` value by using the syntax `comment?: string?` or `comment?: nil | string`.
 
 #### Union Type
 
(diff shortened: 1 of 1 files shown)


BlockLucas:
Hi!

Before in 1.0 there wasn’t any definition for this nilable sugar for references, just a comment in an example ([here](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md/#nil-type)) for a scalar & a definition for properties ([here](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md/#definitions-and-terminology)).
For us is not a PATCH change.

To discuss a little further on this, as we previously talked a little in https://github.com/raml-org/raml-tck/issues/138 with
 this example:
```yaml
types:
  TypeWithTrailing?:
    type: string
  TypeWithTrailing:
    type: int
  TestObject:
    type: object
    properties:
      prop: TypeWithTrailing?
```

If "prop" should reference type "TypeWithTrailing" (int), marking it as nilable.
How we should behave in cases like these:

```yaml
types:
  TypeWithTrailing?:
    type: string
  TestObject:
    type: object
    properties:
      notnilable: TypeWithTrailing?
```
Under that logic, this should fail (because "TypeWithTrailing" doesn't exists), but "TypeWithTrailing?" becomes unreachable or unusable for this sugar

jstoiko:
This was an obvious omission. This has also been extensively discussed/argued before (see #569, #665). 

To answer your question Re: what happens when a type name contains a trailing question mark, I think the referencing mechanism should first look at types with that name before inferring that it means `| nil`. The [`Property Declarations`](https://github.com/raml-org/raml-spec/blob/900c2bb0fadbb487876973daef9960d32f4d44ff/versions/raml-10/raml-10.md#property-declarations) has an example of a "double question mark" key. Maybe we can add one for values too.

versions/raml-10/raml-10.md
```diff
@@ -1036,6 +1036,8 @@ types:
 
 In RAML, the type `nil` is a scalar type that allows only nil data values. Specifically, in YAML it allows only YAML's `null` (or its equivalent representations, such as `~`), in JSON it allows only JSON's `null`, and in XML it allows only XML's `xsi:nil`. In headers, URI parameters, and query parameters, the `nil` type only allows the string value "nil" (case-sensitive); and in turn an instance having the string value "nil" (case-sensitive), when described with the `nil` type, deserializes to a nil value.
 
+Using the type `nil` in a union, makes a type definition nilable, meaning that any instance of that union MAY be a nil data value. When such union is composed of only one type in addition to `| nil`, a trailing question mark `?` in lieu of such union is equivalent. The use of that equivalent alternative syntax SHALL be restricted to [scalar types](#scalar-types) and references to user-defined types, and SHALL NOT be used in [type expressions](#type-expressions).
```
dukesphere, line 1039:
```suggestion
Using the type `nil` in a union makes a type definition nilable, which means that any instance of that union MAY be a nil data value. When such a union is composed of only one type in addition to `| nil`, use of a trailing question mark `?` instead of the union syntax is equivalent. The use of that equivalent, alternative syntax SHALL be restricted to [scalar types](#scalar-types) and references to user-defined types, and SHALL NOT be used in [type expressions](#type-expressions).
```

versions/raml-10/raml-10.md
```diff
@@ -1073,14 +1075,16 @@ types:
...
-                             # comment: string?
+                            # comment: string?
     example:
       name: Fred
       comment: # Providing a value or not providing a value here is allowed.
 ```
 
 Declaring the type of a property to be `nil` represents the lack of a value in a type instance. In a RAML context that requires *values* of type `nil` (vs just type declarations), the usual YAML `null` is used, e.g. when the type is `nil | number` you may use `enum: [ 1, 2, ~ ]` or more explicitly/verbosely `enum: [ 1, 2, !!null "" ]`; in non-inline notation you can just omit the value completely, of course.
 
+Nilable values are not the same as optional properties, e.g. to have a `comment` property that is optional and which can also have a `nil` value, you could define it as `comment?: string?` or `comment?: nil | string`.
```
dukesphere, line 1086: Note that many people don't know the difference bt "e.g." and "i.e.". In general, try to avoid Latin abbreviations.  

```suggestion
Nilable values are not the same as optional properties. For example, you can define a `comment` property that is optional and that accepts a `nil` value by using the syntax `comment?: string?` or `comment?: nil | string`.
```

dukesphere (commented):
suggests avoiding use of terms like "in lieu of" and Latin abbreviations so that it is understood by a wider audience

versions/raml-10/raml-10.md
jstoiko: Noted :)

dukesphere (approved):

PR #735, opened 2020-08-17
Clarify type inference

### LINKED ISSUE #710: Clarify how to infer/determine the default type for ALL built-in types
In the [`Determine Default Types`](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#determine-default-types) section, a number of rules are enumerated to help infer a type without having to specify the `type` facet. 

The first rule mentions how setting a facet that is unique to a given type removes the need to specify the `type` facet. The issue is that it only specifies `properties` as inferring `type: object`. This was clearly an oversight when writing the Spec as there are other types that have facets unique to them such as `fileTypes` for `type: file` and `items` for `type: array`.

### Proposed clarification:
- change the first rule from:
> If, and only if, a type declaration contains a properties facet, then the default type is object

to:
> If, and only if, a type declaration contains a facet that is unique to that type, then the default type is that type
- adapt subsequent rules referring to the first one
- make it clear that the `type: object`/`properties` in that first rule is only an example

jstoiko:
Closes #710

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1277,7 +1277,7 @@ User-defined facets by definition are not built into this RAML specification, an
 
 A RAML processor must be able to determine the default type of a type declaration by using the following rules:
 
-* If, and only if, a type declaration contains a facet that is unique to that type, then its default type is inferred to be the only one with support for the facet being used.
+* If, and only if, a type declaration contains a facet that is unique to that type, its default type is then inferred to be the only one with support for the facet being used.
 
 For example, if a type declaration contains a `properties` facet, its default type is `object`. The following snippet exemplifies this rule:
 
@@ -1289,7 +1289,7 @@ types:
     properties:
 ```
 
-* If, and only if, a type declaration contains neither a facet that is unique to a given type as described in the previous rule above, nor a `type` or `schema` facet, then the default type is `string`. The following snippet exemplifies this rule:
+* If, and only if, a type declaration contains a facet that is neither unique to a given type, as described in the previous rule above, nor a `type` or `schema` facet, then the default type is `string`. The following snippet exemplifies this rule:
 
 ```yaml
 types:
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1277,25 +1277,19 @@ User-defined facets by definition are not built into this RAML specification, an
 
 A RAML processor must be able to determine the default type of a type declaration by using the following rules:
 
-* If, and only if, a type declaration contains a `properties` facet, then the default type is `object`. The following snippet exemplifies this rule:
+* If, and only if, a type declaration contains a facet that is unique to that type, then its default type is inferred to be the only one with support for the facet being used.
+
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -1277,25 +1277,19 @@ User-defined facets by definition are not built into this RAML specification, an
 
 A RAML processor must be able to determine the default type of a type declaration by using the following rules:
 
-* If, and only if, a type declaration contains a `properties` facet, then the default type is `object`. The following snippet exemplifies this rule:
+* If, and only if, a type declaration contains a facet that is unique to that type, then its default type is inferred to be the only one with support for the facet being used.
```
priyaayyar, line 1280:
```suggestion
* If, and only if, a type declaration contains a facet that is unique to that type, its default type is then inferred to be the only one with support for the facet being used.
```

versions/raml-10/raml-10.md
```diff
@@ -1277,25 +1277,19 @@ User-defined facets by definition are not built into this RAML specification, an
...
-types:
-  Person:
-    type: object
-    properties:
-```
-
-This rule can also be written as follows:
+For example, if a type declaration contains a `properties` facet, its default type is `object`. The following snippet exemplifies this rule:
 
 ```yaml
```
priyaayyar, line 1284: Put the code in a code tag?

versions/raml-10/raml-10.md
```diff
@@ -1277,25 +1277,19 @@ User-defined facets by definition are not built into this RAML specification, an
...
 types:
   Person:
-    # default type is `object`, no need to explicitly define it
+    # default type is "object" because "properties" is unique to that type
+    # i.e. no need to explicitly define it, "type: object" is inferred
     properties:
 ```
 
-* If, and only if, a type declaration contains neither a `properties` facet nor a `type` or `schema` facet, then the default type is `string`. The following snippet exemplifies this rule:
+* If, and only if, a type declaration contains neither a facet that is unique to a given type as described in the previous rule above, nor a `type` or `schema` facet, then the default type is `string`. The following snippet exemplifies this rule:
```
priyaayyar, line 1292:
```suggestion
* If, and only if, a type declaration contains a facet that is neither unique to a given type, as described in the previous rule above, nor a `type` or `schema` facet, then the default type is `string`. The following snippet exemplifies this rule:
```

versions/raml-10/raml-10.md
```diff
@@ -1277,25 +1277,19 @@ User-defined facets by definition are not built into this RAML specification, an
...
 
 ```yaml
```
priyaayyar, line 1294: Put the code in a code tag?

priyaayyar (changes requested):

versions/raml-10/raml-10.md
jstoiko: That's the case, i.e. that's what those 3 backticks do...

dukesphere (approved):

priyaayyar (approved):

PR #742, opened 2020-08-17
Clarify relative paths in overlays/extensions

### LINKED ISSUE #719: Clarify how to resolve `!include`s and libraries inside Overlay and Extension fragments
In the context of RAML Overlay and Extension fragments, it is not clear whether the root is the overlay/extension fragment itself or the RAML file that it extends. This matters when an overlay or extension fragment is not located in the same folder as the RAML file that it extends.

### Proposed changes
At the end of the following paragraph:
> The remainder of an overlay or extension document follows the same rules as a RAML API definition, but with certain restrictions in case of an overlay.

(see ["Overlays and Extensions"](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#overlays-and-extensions))

Add the following clarification: `The resolution of any relative or absolute paths in an overlay or extension document such as any !include tags or libraries MUST be done relatively to such overlay or extension document.`

jstoiko:
Closes #719


versions/raml-10/raml-10.md
```diff
@@ -3314,7 +3314,7 @@ API definitions might need to be extended in a variety of ways for different nee
...
 
-The remainder of an overlay or extension document follows the same rules as a RAML API definition, but with certain [restrictions](#overlays) in case of an overlay.
+The remainder of an overlay or extension document follows the same rules as a RAML API definition, but with certain [restrictions](#overlays) in case of an overlay. The resolution of paths in an overlay or extension document, such as those for `!include` tags or libraries, MUST be relative to the document from which the reference is being made, as opposed to the API root document.
```
priyaayyar, line 3317:
```suggestion
The remaining overlay or extension document follows the same rules as the RAML API definition, but with certain [restrictions](#overlays) in case of an overlay. The resolution of paths in an overlay or extension document, such as those for `!include` tags or libraries, must be relative to the document from which the reference is being made, as opposed to the API root document.
```

priyaayyar (changes requested):

versions/raml-10/raml-10.md
dukesphere: should stay "MUST" not lowercase "must"

dukesphere (approved):
I don't think the suggested change is necessary. Approving as-is.

versions/raml-10/raml-10.md
dukesphere, line 3317:
```suggestion
The remainder of an overlay or extension document follows the same rules as a RAML API definition, but with certain [restrictions](#overlays) in case of an overlay. The resolution of paths in an overlay or extension document, such as those for `!include` tags or libraries, MUST be relative to the document from which the reference is made, as opposed to the API root document.
```

dukesphere (commented):
small change, restores MUST, as well

priyaayyar (approved):

PR #744, opened 2020-08-17
Clarify value type of the `is` node

### LINKED ISSUE #724: Clarify that value of `is` in Applying Traits MUST be array
In section ["Applying Resource Types and Traits"](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#applying-resource-types-and-traits), it should be explicit that the value of `is` MUST be an array.

### Proposed Change

Change:

> Similarly, a method can specify one or more traits it inherits using the OPTIONAL is node. The value of a trait is an array (...)

to:

> Similarly, a method can specify one or more traits it inherits using the OPTIONAL is node. The value of a trait MUST be an array (...)

jstoiko:
Closes #724

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -2200,7 +2200,7 @@ resourceTypes:
 
 A resource can specify the resource type from which it is derived using the OPTIONAL **type** node. The value MUST be the name of a resource type defined within the root-level resourceTypes node or in a library. Resource type definitions do not apply to existing nested resources.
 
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -2200,7 +2200,7 @@ resourceTypes:
...
+Similarly, a method can specify one or more traits it inherits using the OPTIONAL **is** node. The value of a trait MUST be an array of any number of elements where each MUST be the name of a trait defined within the root-level traits node or in a library. A trait can also be applied to a resource by using the **is** node. Using this node is equivalent to applying the trait to all methods for that resource, whether declared explicitly in the resource definition or inherited from a resource type. A trait is applied to a method in left-to-right order, according to the traits defined in the **is** node. Trait definitions do not apply to nested resources.
```
dukesphere, line 2203:
```suggestion
Similarly, a method can specify one or more traits it inherits by using the OPTIONAL **is** node. The value of a trait MUST be an array of any number of elements. Each element MUST be the name of a trait defined within the root-level traits node or in a library. A trait can also be applied to a resource by using the **is** node. Using this node is equivalent to applying the trait to all methods for that resource, whether the method is declared explicitly in the resource definition or inherited from a resource type. A trait is applied to a method in left-to-right order, according to the traits defined in the **is** node. Trait definitions do not apply to nested resources.
```

dukesphere (changes requested):
clarification

dukesphere (approved):

PR #745, opened 2020-08-17
Clarify usage of the "NamedExample" typed fragment

### LINKED ISSUE #727: Clarify "NamedExample" typed fragment description and usage
The `NamedExample` fragment, being in the singular form, may lead to confusion as to whether it could be used to define a singular example as with `example` node, or any number of examples as with `examples` node. The ["Typed Fragments"](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#typed-fragments) section unambiguously refers to `Examples` as being the "Relevant RAML Specification Section" for this type of  fragment, however the description should be clarified.
 
### Proposed changes
- In the table of the ["Typed Fragments"](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#typed-fragments), change:
> A declaration of the examples facet, whose key is a name of an example and whose value describes the example

to:
> A declaration of the examples facet, whose keys are names of examples and whose values describe each of the examples
- Add a RAML example making of `NamedExample` at the bottom of that same section

jstoiko:
Closes #727

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -3065,7 +3065,7 @@ A file to be included MAY begin with a RAML fragment identifier line, which cons
 |:-------------------|:------------| :-----------------------------------|
 | DocumentationItem | An item in the collection of items that is the value of the root-level documentation node | [User Documentation](#user-documentation)
 | DataType | A data type declaration where the type node may be used | [Types](#defining-types)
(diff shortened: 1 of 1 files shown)


dukesphere (changes requested):
clarification

dukesphere (approved):

PR #747, opened 2020-08-17
Clarify use of traits and resource types parameters in modularization

### LINKED ISSUE #730: Clarify whether resource type and trait parameters can be used in modularization
The ["Resource Type and Trait Parameters" section](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#resource-type-and-trait-parameters) contains a restriction on the use of parameters but only includes are mentioned. This restriction should apply to all types of modularization.

### Proposed Change

Rewrite:
> Parameters cannot be used within an !include tag specification of the include file location.

to read:
> Parameters cannot be used within any file location used in the context of [modularization](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#modularization), that is any file location defined in the `!include` tag or as value of any of the `uses` and `extends` nodes.

jstoiko:
Closes #730

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -2331,7 +2331,7 @@ traits:
         example: <<methodName>>=h8duh3uhhu38   # e.g. get=h8duh3uhhu38
 ```
 
-Parameters cannot be used within any file location used in the context of [modularization](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#modularization), that is any file location defined in the `!include` tag or as value of any of the `uses` and `extends` nodes.
+Parameters cannot be used within any file location that is used in the context of [modularization](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#modularization), that is, any file location defined in the `!include` tag or as a value of any of the `uses` or `extends` nodes.
 
 ### Declaring HTTP Methods as Optional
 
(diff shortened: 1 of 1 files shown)


versions/raml-10/raml-10.md
```diff
@@ -2331,7 +2331,7 @@ traits:
         example: <<methodName>>=h8duh3uhhu38   # e.g. get=h8duh3uhhu38
 ```
 
-Parameters cannot be used within an !include tag specification of the include file location.
+Parameters cannot be used within any file location used in the context of [modularization](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#modularization), that is any file location defined in the `!include` tag or as value of any of the `uses` and `extends` nodes.
```
dukesphere, line 2334:
```suggestion
Parameters cannot be used within any file location that is used in the context of [modularization](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#modularization), that is, any file location defined in the `!include` tag or as value of any of the `uses` or `extends` nodes.
```

dukesphere (commented):
suggested tweak

dukesphere (commented):
small tweaks. changed an "and"  to "or" since appears that it would apply to either case (`uses` or `extends` nodes).

dukesphere (changes requested):
requested changes

dukesphere (approved):

PR #755, opened 2020-11-24
Clarify how to process JSON-formatted data structures without a $schema key

### LINKED ISSUE #751: Json Schema draft compatibility limit
Hi! is there an upper and lower bound of json schema draft versions that can be included in a RAML?

The only reference there is to a JSON Schema Draft number is in the "references" section but it never explicitly states if RAML 1.0 supports a limited set of drafts.

Thanks!

jstoiko:
As discussed in #751.

### DIFF
--- a/versions/raml-10/raml-10.md
+++ b/versions/raml-10/raml-10.md
@@ -1226,6 +1226,8 @@ types:
 
 A RAML processor MUST be able to interpret and apply JSON Schema and XML Schema.
 
+Because JSON is also a valid YAML 1.2 syntax, a RAML processor SHOULD interpret JSON-formatted data structures that do not contain a `$schema` key as RAML type declarations.
+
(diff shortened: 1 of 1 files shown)


dukesphere (approved):
looks good

GauthierPLM:
I think adding a link to a reference explaining why `JSON is also a valid YAML 1.2` would be helpful for readers. People might have not heard of this before and not understand it without additional research. Having a reference would ensure this is sourced and indeed a spec the reader can refer to.

FYI, I never heard about this before and discovered while reading this PR.

jstoiko:
Thanks for the note @GauthierPLM. It's in the [YAML 1.2 spec](https://yaml.org/spec/1.2/spec.html), both in the initial introduction and "1.3 Relation to JSON" sections.

Issue #767, opened 2022-02-14
Ambiguous/wrong example of union in Query String section


looseale:
Hi.
I'm reporting an error in an example in the spec of RAML 1.0.
In the section [The Query String as a Whole](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#the-query-string-as-a-whole) there is an example API which comments and examples are showing that an union elements must be validated exclusively (like a JSON Schema oneOf).

```
#%RAML 1.0
title: Illustrate query parameter variations
types:
  lat-long: # lat & long required; mutually exclusive with location
    properties:
      lat: number
      long: number
  loc: # location required; mutually exclusive with lat & long
    properties:
      location:
  paging: # each is optional, not exclusive with anything
    properties:
      start?: number
      page-size?: number
/locations:
  get:
    queryString:
      type: [paging,  lat-long | loc ]
      examples:
        first:
          value:
            start: 2
            lat: 12
            long: 13
        second:
          value:
            start: 2
            page-size: 20
            location: 1,2
        third:  # not valid
          value:
            lat: 12
            location: 2
          strict: false # because it's not valid
```

But, in the section [Union Type](https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md#union-type), the text express clearly that they could be inclusive (like a JSON Schema anyOf): `An instance of a union type SHALL be considered valid if and only if it meets all restrictions associated with at least one of the super types`. It also give an example very similar to the one in Query String section, multiple inheritance mixed with unions.

uip-robot-zz:
Git2Gus App is installed but the `.git2gus/config.json` doesn't exist.

PR #768, opened 2022-02-16
#767 - Correct and simplify query string example

looseale:
This example shows that unions work like exclusive (gives the idea that an instance must match only with one element of the union) but in the union type section always says that it should match `at least with one element` of the union.


uip-robot-zz:
Git2Gus App is installed but the `.git2gus/config.json` doesn't exist.

PR #769, opened 2022-03-25
Update raml-10.md

antoniogarrote:
Making the declaration of the annotation type optional.
This will allow better interoperability with OAS, where extensions are not required to have a strict schema.
This also makes it possible to easily generate annotations when enriching the spec via software for automation use cases.
The change is backward compatible. Users that want to provide a better experience can still provide the annotation type.


uip-robot-zz:
Git2Gus App is installed but the `.git2gus/config.json` doesn't exist.

